PR #216, opened 2020-06-23
Modify divider control such as psd

javadsamiee:
Add active sate for line and drop show options
Some bug fix


Naaaaiix (changes requested):
1. On 'src/blocks/divider-maxi/inspector.js', `getHorizontalOptions` and `getVerticalOptions` are not necesary anymore, even `getLinesQuantity` function.
2. On 'src/blocks/divider-maxi/edit.js', element for line 2 still exists.
3. Line defaults has strange behaivour:
line defaults

Naaaaiix (approved):

PR #320, opened 2020-07-21
bg not working as they should

javadsamiee:
bg not working as they should


Naaaaiix (changes requested):
IMO this PR is not a good idea and it breaks with some basic necessities of the plugin. For example, based on Container Maxi example (the one which has ShapeDivider), doesn't have logic that it contains 'height' CSS property to be 350px. Is completely non-viable: 
Captura de pantalla 2020-07-22 a las 10 27 55
Also, we have to be really concern with the fact that blocks should be plastics and adapt to the content and necessities, and the most important part: what you see on the backend should be seen on frontend. This is how it looks on frontend:
Captura de pantalla 2020-07-22 a las 10 31 36
Please, check [this alternative branch](https://github.com/yeahcan/maxi-blocks/compare/joshua-bg_not_working_as_they_should_alternative) that uses a much easier method to fix the problem and let me know if it covers the necessities of the issue 👍

Naaaaiix (approved):

PR #327, opened 2020-07-22
First prototype for ClipPathControl

Naaaaiix:
## Reproduce
1. Create an Image Block
2. Get to Advanced tab and play with clip-path

## Known problems
There's is no visual part, is just on settings sidebar.

## Information
In order to check it, you can use [clippy](https://bennettfeely.com/clippy/) to create the shape, and after, taking the numbers from the bottom, reproduce them into clip-path options.


elzadj (approved):
Looks great, need to send Marko to test after :+1:

PR #332, opened 2020-07-24
Alignment fixes - for Image Block's caption and all text hover animations

elzadj:
#323


Naaaaiix (approved):
Needs minor fixes. Going to create an alternative branch in order to fix them and to push them into master 👍

PR #337, opened 2020-07-27
Javad new design for viewport of motion effects

javadsamiee:
New design for viewport of Motion Effects


Naaaaiix (commented):
Good job buddy! I love this AdvancedRangeControl ❤️

PR #338, opened 2020-07-28
Arrow on Container Maxi & Responsive values

Naaaaiix:
This PR tries to solve two different problems:
1. #329:
    - To reproduce:
    1. Create a Container Maxi block as first element on editor
    2. Try ArrowControl component on settings sidebar
    3. Create a Container Maxi block inside a Container Maxi or a Column Maxi
    4. Try ArrowControl component on settings sidebar

2. #336:
    - To reproduce:
    1. Create a Maxi Block
    2. Set a value in one responsive stage
    3. Change the responsive stage to one smaller
    4. Check if the value on smaller responsive stage is same than in the stage were values has been set


javadsamiee (approved):
good job man, your code was awesome as always and I love to transform component that is powerful and useful

PR #342, opened 2020-07-28
Clip-path background compatibility - #316

Naaaaiix:
Related with #316 

To reproduce:
1. Create a block (not Button Maxi)
2. Set a background
3. Check ClipPathControl


javadsamiee (commented):
Have you deleted videoPlayer component?

PR #350, opened 2020-08-03
Javad add reset button and defaults to motion effects

javadsamiee:


Naaaaiix (changes requested):
May I suggest to create an 'object.json' file on 'motion-control' folder to understand better the required object for the component. Know is on 'default.js' on 'style' folder; just to follow the rest of components.

Minor things, but for the rest really good job. User experience for Motion Effects is really good, works smothly and code is clean and good structured. Congrats men! 😄

Naaaaiix (approved):

PR #356, opened 2020-08-07
Javad hover animation problem

javadsamiee:


Naaaaiix (approved):
Amazing job, stunning! 💪 😄 
Couple of questions:
1. Is just on Image Maxi; was some kind of prototype to test before implementing to rest?
2. Are you going to abstract '.maxi-hover-details' element? In that case, maybe we can thing on creating an element that wraps BackgroundDisplayer and this one.
3. This is for the future, when optimizing&refactoring: instead of adding the motion-effect properties or the hover properties as 'data' attributes into elements (that is not bad), may we can extend the API to save that properties and after deliver them into frontend in 'head' element. I think is cleaner, and our clients who check the code won't see that. Also will give some fasten to our plugin, because we won't need to check elements looking for a concrete effect: we'll be able to request a function to do it directly.
4. Just seen: 'trim()' values on 'inputs', or in values as 'SizeControl'. If not, they can have a 0 in front.

Thanks @javadsamiee 👍

Naaaaiix (approved):
Lovely!

PR #373, opened 2020-08-14
Setting default values on reset

Naaaaiix:
This PR contains 3 fixes, 1 big and 2 minor:
1. Fix all reset buttons comparing with default values as commented on #128.
2. Standardize props names on components
3. Fix all instances of DefaultStylesControl as commented on #363

## To reproduce
For first fix:
1. Create a block
2. Change default value in setting
3. Press reset button

For third fix:
1. Create a block
2. Click default on Border, Text Shadow or Box Shadow.
3. Modify any value or press reset button.

## Known errors
As described #372, RangeControl presents some incompatibilities when is used without reset button. This makes that in places like 'font-size' in TypographyControl, when pressing reset, range is not moving (value is correct, just a UX error).


javadsamiee (commented):
First one: after a click on reset button on typography section, input value changes correctly but slider remaining
90250733-c827f300-de51-11ea-9818-993ecd5a550c

PR #381, opened 2020-08-19
Javad text headings enhance

javadsamiee:


Naaaaiix (changes requested):
Other issues:
1. How can I transform a list to paragraph? So, I can go from paragraph to list, but not from list to paragraph.
2. On pasting content, there's a single Text Maxi block that remains in the top:
textpastingerror
3. On splitting a list, the blocks duplicate with an empty block first:
textsplitlisterror
4. On setting bold/italic/link inside a list, the format breaks down:
textboldlisterror

PR #386, opened 2020-08-20
Fix Style and image

javadsamiee:


Naaaaiix (commented):
IMO we should avoid css pseudoelements to load files and images, they are cause of lots of problems. Don't you think it would be better to modify './js/gutenberg-ui.js:20' to be like:
`</i> ${ __( 'Maxi Cloud Library', 'maxi-blocks') }</a>`
See the difference is in 'src' attribute that starts with `../` instead of `/`.

PR #393, opened 2020-08-25
ESLint, Prettier & AirBnB - Update

Naaaaiix:
Hey there devs!
Here you have a version with next changes:
1. Installed the dependencies ESLint, Prettier & AirBnB
2. Updated plugin dependencies

# IMPORTANT
Code has been linted but not fixed. It throws errors of code everywhere, alarms and even nuclear weapons, but when the code is mounted on build the plugin runs perfectly. Also, the plan is not to merge this branch directly to master because fixing all this errors and then merging from other branches could be totally insane. The goal of this branch is to have a base and check if the linter works as we expect. Then we can take the config and put in the master when everybody have merged the branch into master.

Other IMPORTANT thing: as is an update and you'll require to install new and update dependencies, I recommend you to install this branch in a completely different folder. Not update in your normal plugin folder environment if you don't want to downgrade after when coming back to master or to any other branch.

There are some ideas I really liked. For example, instead using ternary conditionals with `? :` we can use as @javadsamiee used sometimes `&& ||`; that I liked.

I expect your comments and suggestions about this implementation. Thanks! 😄


javadsamiee (commented):
I have tested it and I think it is incredible so we can write code without anxious and we will have clean codes everywhere.

Good job man, thanks for doing it, we need it.

P.S We should always follow the standards to prevent problems in the future.

PR #397, opened 2020-09-01
Svitlana cloud library

elzadj:
Add SVG icon Maxi and components / settings for it
Remove old Icon Maxi


Naaaaiix (approved):
Some things I've found:
1. Alignment is not working correctly:
svgmaxialignment

2. When restarting size is oversizing:
svgmaxiresetsize
3. Not sure if is a bug or what hapend: Image Maxi block has compatibility with SVG. Anyway, it doesn't work when setting a SVG on Image Maxi.

Some question:
1. If users would like to change the icon, should they remove the block and create another one?

Anyway, good job! The Library for picking up the icons looks really cool, smoothly and easy to use. Major of my comments on code are because there are some things not necessary or some functions that needs to be refactored, but the result anyway is really good 😄 
Approving the branch for being merged, feel free to do it onces the changes are done or if you consider 👍

Naaaaiix (approved):
There are still some things to clean but I guess this weekend, implementing the linter, it will look better. Merging 😄

PR #403, opened 2020-09-03
issue 24

javadsamiee:
Add None option for Background
Move overlay settings from Background to new Overlay section


Naaaaiix (approved):
It works perfect, and code is clean; amazing job!

I really like the concept of the 'activeMedia', good idea @javadsamiee 

What I'm not sure is why do why need this overlay? Is for having an image with a color or a gradient on it? Let's be sure the result we are looking for this is not possible to do it with another existing options because, if not, we can find ourselves with redundant settings when we have a lot of them on sidebar and users will get overload and confused 👍 

I approve, but before merging, let's talk about it 😸

Naaaaiix (commented):
Still finding the error with the hovering. For the rest is perfect 👍

PR #413, opened 2020-09-07
issue 51

javadsamiee:
Fix Display Block radio option style
Fix Transform styles

### DIFF
--- a/src/components/display-control/index.js
+++ b/src/components/display-control/index.js
@@ -65,16 +65,18 @@ const DisplayControl = props => {
 		];
 	};
 	return (
-		<div className={classes}>
-			<RadioControl
-				label={__('Display block', 'maxi-blocks')}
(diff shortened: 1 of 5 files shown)


Naaaaiix (commented):
Perfect coding, didn't find any mistake. Also the UI is clean and really beautiful. Good job buddy!

Some UX bugs and suggestions:
1. After reseting Translate, the mover display doesn't come back to it's default place and is not moving anymore. Also you have a console error:
transform-translate

2. On Origin, when default, there's no circle marked. After you mark, when clicking reset, the last circle keeps marked even when it returns to default value. Also, I miss seeing what's the unit for the value: is 'px', '%'?
transform-origin

Naaaaiix (commented):
Great job buddy, just one little thing and we are done with it! 😄

PR #424, opened 2020-09-10
Row Snapping Feature 

abder:
This pull request presents the feature of row responsive snapping but it still needs extensive testing before merging, which what I need your help with guys. 

Please check it out and comment any issues and suggestions 

Thanks!


javadsamiee (commented):
Good job man for your hard work.

That was good,  I think we need to add an active class for the current selected item because there are lots of items, and finding the current item is hard.

A question: Did you do for responsive modes? I think it's working on a desktop

Naaaaiix (changes requested):
Good job men, is a big one this!

Naaaaiix (changes requested):
Much better!

Naaaaiix (approved):
Much better!

PR #432, opened 2020-09-14
Issue 55

javadsamiee:
https://github.com/yeahcan/maxi-blocks/issues/415#issuecomment-690908241

Add Orange icons to padding in the toolbar
Fix Column size input style in the toolbar

### DIFF
--- a/src/components/toolbar/components/padding-margin/index.js
+++ b/src/components/toolbar/components/padding-margin/index.js
@@ -36,11 +36,6 @@ const PaddingMargin = props => {
 			className='toolbar-item__padding-margin'
 			tooltip={__('Padding & Margin', 'maxi-blocks')}
 			icon={toolbarPadding}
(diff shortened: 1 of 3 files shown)


Naaaaiix (approved):
Great job, works easily and smoothly 😄 

I would like to say that I'm not sure about this improvement for the next reason: I'm not quite sure what till be the paper of the margin on the Column Maxi. Margin was not appearing on the toolbar because in some moment we made Column doesn't have margin option, but after we put it again. In case we'll use margin on Column Maxi I would suggest that, more than putting the orange button to make the user go to settings sidebar, margin options should appear in the Toolbar as it happens with the rest of blocks; this would be IMO the way to keep the coherence with the rest of the blocks 👍

Naaaaiix (approved):
Sublime, merging 👍

PR #434, opened 2020-09-15
Issue 57

javadsamiee:
Fix full-width option style to fancy for all blocks


Naaaaiix (approved):
Good job @javadsamiee , well done!

I would suggest that more than employing time doing the change for a RadioControl with the class 'maxi-fancy-radio-control', abstract the component and implement it; at least in the new instances. If not, we'll do double work when we will abstract it: now adapting with this rudimentary system, and after adapting for the new component 👍 Anyway this is done, so merging and fixing in future 😄

PR #435, opened 2020-09-15
Issue 58

javadsamiee:
Change stretch to flex-start in col and row to fix button stretch problem

### DIFF
--- a/src/blocks/row-maxi/attributes.js
+++ b/src/blocks/row-maxi/attributes.js
@@ -16,7 +16,7 @@ const attributes = {
 	},
 	verticalAlign: {
 		type: 'string',
-		default: 'stretch',
+		default: 'flex-start',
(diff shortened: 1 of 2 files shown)


Naaaaiix (approved):
Good seeing, but I would like to discuss a bit about it because I'm not sure what would be better for users 🤔 :
The problem is not just with button, what it was happening with it was an 'expected' behavior. What is happening is that with 'stretch' value for 'vertical-align' all columns get the same height automatically like this image:
rowverticalalign

In case we change for 'flex-start' as you suggest on this PR, the column height remains in its content:
Captura de pantalla 2020-09-15 a las 12 03 30

IMO 'flex-start' is a good choice for users, what do you think?

Anyway, the change is approved and can be merged in any moment 👍

PR #489, opened 2020-09-16
Issue 60

javadsamiee:
Add round to fix opacity range control, Thanks Josh for the 'round' suggestion 😸


Naaaaiix (approved):
A pleasure, as always! 😄

PR #496, opened 2020-09-16
Issue 65

javadsamiee:
Create new component - ImageAltControl


Naaaaiix (approved):
Well done buddy! 👍 

I would suggest to set Image Alt Tag inside an accordion as the rest of options; IMO in this way it breaks the harmony and generates a big difference with the rest of the blocks. What do you think?

Naaaaiix (approved):
Perfect! 😄

PR #501, opened 2020-09-18
Issue 70

javadsamiee:
Add move up and down for blocks
Redesign and change icons or move right and left columns


Naaaaiix (changes requested):
Good job men! Couple of comments related with UX:

1- Buttons are small, but the cursor invites you to click even out of the button. You expect it to move, but it doesn't.
Captura de pantalla 2020-09-20 a las 20 06 55
Captura de pantalla 2020-09-20 a las 20 05 00

2. I would suggest that, in case is not possible to move to some direction, button change color to advice users that this movement is not possible.

Naaaaiix (approved):
Fantastic job! Love it 😍

PR #502, opened 2020-09-21
Issue 71

javadsamiee:
Minor fix

### DIFF
--- a/src/blocks/divider-maxi/data.js
+++ b/src/blocks/divider-maxi/data.js
@@ -10,7 +10,7 @@ const size = {
 		'max-heightUnit': 'px',
 		'max-height': '',
 		heightUnit: 'px',
-		height: 20,
+		height: 100,
 		'min-heightUnit': 'px',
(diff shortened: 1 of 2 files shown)


Naaaaiix (approved):
Looks better, good job 😄 

What I found is that is imposible to change the height when modifying any 'Width/Height' value:
dividersizing

Naaaaiix (commented):
Good step, but I'm finding other errors. Anyway, I think they are not related with this PR, so may we should abstract new issues for each one:

1. Divider Maxi is not centred:
Captura de pantalla 2020-09-23 a las 9 28 17

2. Not sure if FullSizeControl or SizeControl (FullSizeControl is made with different SizeControls) have an error and is not displaying the correct value:
Captura de pantalla 2020-09-23 a las 9 35 33
Maybe related with #458 🤔 

3. On changing the height from settings sidebar is affecting the inner element of the block on the editor, and not the resizer box. The result is this:
dividerheighterror

PR #506, opened 2020-09-22
Issue 74

javadsamiee:
Change Width/Height label to Image Dimension
Fix Width/Height

### DIFF
--- a/src/blocks/image-maxi/edit.js
+++ b/src/blocks/image-maxi/edit.js
@@ -76,6 +76,7 @@ class edit extends MaxiBlock {
 
 	get getNormalObject() {
 		const {
+			size,
 			alignment,
 			opacity,
 			boxShadow,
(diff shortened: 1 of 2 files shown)


Naaaaiix (changes requested):
Mayday, mayday! Control, we have a problem! lol
imagesizeerror

Naaaaiix (approved):
Perfect, works much better! 😄 

Not sure if is related with this PR, maybe we need to merge a new issue, but due to the element `<div class="maxi-block-hover-element"></div>` the image doesn't get the proper alignment on the frontend:
imagesizeerror

Also, the element `<div class="maxi-hover-details"></div>` is in front of the image and when selecting from console appears first than the image. 

Approving the PR, merge yourself @javadsamiee onces you've read my comment 🙏

PR #507, opened 2020-09-22
Issue 75

javadsamiee:
Some fix

### DIFF
--- a/src/blocks/divider-maxi/edit.js
+++ b/src/blocks/divider-maxi/edit.js
@@ -37,8 +37,9 @@ class edit extends MaxiBlock {
 
 		let response = {
 			[uniqueID]: this.getNormalObject,
-			[`${uniqueID}:hover`]: this.getHoverObject,
-			[`${uniqueID} > hr.maxi-divider-block__divider`]: this
+			[`${uniqueID}:hover hr.maxi-divider-block__divider`]: this
+				.getHoverObject,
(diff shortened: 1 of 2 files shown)


Naaaaiix (changes requested):
Finding some errors:
1. The default 'opacityHover' object should have an empty value, if not is 'active' and will affect the block. That's the reason why the opacity on 'hover' appears as 100%. [This is the line](https://github.com/yeahcan/maxi-blocks/blob/dffb64f41fbed909499ebd0a17f55626d9f21da7/src/extensions/styles/defaults.js#L54) that should be modified: '1' for an empty value.

2. Opacity element on [`getHoverObject`](https://github.com/yeahcan/maxi-blocks/blob/a4d4b42573153c39e5bd7fddaea90dd410f12932/src/blocks/divider-maxi/edit.js#L105) is not in the correct place, IMO. In the normal version, not hover, opacity is affecting the hole block; in the hover version is affecting only the line. So, what is happening? Hover is not working correctly because is targeting two different elements:
dividertargets

Also, when dragging with the Divider Maxi, after, the block is not selected.

Naaaaiix (approved):
Box-shadow is working good now, but the other 2 errors commented on the last review are still there. Decide if you want to fix on this branch or generate new issues and fix them apart. Shouldn't be a big deal 👍

PR #508, opened 2020-09-22
Fix Issue 436

abder:
This PR aims to fix the issue explained on #436 

captured (7)

### DIFF
--- a/src/blocks/container-maxi/inspector.js
+++ b/src/blocks/container-maxi/inspector.js
@@ -416,6 +416,7 @@ const Inspector = props => {
 																		breakpoint={
 																			deviceType
 																		}
+																		displayToggle
 																	/>
 																	<BackgroundControl
 																		background={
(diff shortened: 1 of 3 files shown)


Naaaaiix (commented):
First of all, good job @DOUARA ; you used some good details, and looks you're getting into the dynamics, congrats 😸 

About the PR itself: agree with @javadsamiee, the plan was to integrate it into the hover options to toggle them. But at same time I understand @DOUARA confusion for it: the problem, IMO, is that we are mixing on the same settings sidebar many options as Opacity and Background. So, as we talked on the Scrum meeting, let's stop this and find other ways to do it 👍

javadsamiee (commented):
@Naaaaiix I think because we want to fix this issue completely via this issue https://github.com/yeahcan/maxi-blocks/issues/514 so we can skip this PR and also close this issue https://github.com/yeahcan/maxi-blocks/issues/436

PR #515, opened 2020-09-23
issue-511

abder:
The issue was that the **resizer** changes it's width slightly when we change the height with it, to fix this I had to set 100% width to it  with !important to override the inline style :smile: 

captured (3)


javadsamiee (commented):
Good job, apparently it's working correctly

PR #525, opened 2020-09-25
issue-455

abder:
This PR aims to fix the issue described in #455 

I think in the future we need to use our size-control component for it, it looks much better: 
Screenshot from 2020-09-25 09-38-53

The only thing is that wee need to to add the option to it to display a fixed unit

### DIFF
--- a/src/blocks/column-maxi/inspector.js
+++ b/src/blocks/column-maxi/inspector.js
@@ -17,7 +17,6 @@ import {
 	BoxShadowControl,
 	FullSizeControl,
 	SettingTabsControl,
-	SizeControl,
 	__experimentalZIndexControl,
 	__experimentalResponsiveSelector,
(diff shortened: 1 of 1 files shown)


Naaaaiix (approved):
Completely agree with you, and I think @javadsamiee is in the same path: as soon it will be possible, we should stop using Gutenberg (even in SizeControl) and use our own components to have more freedom to modify them. Time to time my friend, but is good we have all the same objectives 😄 

Fix the minor comment I left and after I'll merge the branch. Good job 👍

javadsamiee (approved):
Pasted_Image_9_25_20__2_03_PM

Please create a new issue for this one, when we have a long title it's better to show in above of range, we can add a new attribute to range control to do it

javadsamiee (approved):

PR #531, opened 2020-09-28
issue-530

abder:
a Quick fix to the issue mentioned in #issue-530


Naaaaiix (commented):
Not sure if it's a correct/necessary fix for 2 reasons:
1. Let's see how fixing #530 main problem affect this, because I think the problem with the heights is because the margin you're setting in.
2. In case is necessary to change the height on the ResizableBox, do it from its inner props 👍 

Let's come back to this PR onces #530 is fixed 🙏

PR #534, opened 2020-09-30
Issue 458

abder:
This PR aims to close the issue in #458 

As we are using size-control component with all the blocks in many different places, I have tested it in all of them, but please do more tests to it to ensure that's working correctly :pray:


javadsamiee (approved):
Good

PR #538, opened 2020-09-30
Issue-523

abder:
Fix to issue mentioned in #523


javadsamiee (commented):
IMO, I think it's better to disable all items under columns when we active handles because we will have an old problem again. please test with a column that has a text and enable handles so try to resize the column.

PR #539, opened 2020-09-30
issue-448

abder:


javadsamiee (commented):
I can not type minus value but I can use arrow up and down (keyboard) to change to a minus value, please this 😃

Naaaaiix (approved):
Perfect, good job 😄

PR #542, opened 2020-09-30
issue-451

abder:
This PR aims to fix the issue mentioned in #451

### DIFF
--- a/src/components/column-pattern/index.js
+++ b/src/components/column-pattern/index.js
@@ -97,7 +97,7 @@ const ColumnPatternsInspector = props => {
 	 * @returns {Array} Array with saved content
 	 */
 	const getCurrentContent = (blockIds, newTemplate = []) => {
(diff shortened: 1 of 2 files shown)


javadsamiee (approved):
Good job 👍

Naaaaiix (commented):
As I understood #451, the plan was to hide the column pattern from the Toolbar when the big pattern selector is on the center of the block for first time; @javadsamiee correct me if I'm wrong.

Another bug found, maybe not related with the issue and the PR, is that if I modify the pattern using the handlers doesn't deactivate the active status, so the pattern selected doesn't fit with the pattern 👍

Naaaaiix (approved):
Perfect, merging! 👍 Good job men

Naaaaiix (changes requested):

Naaaaiix (approved):

PR #545, opened 2020-10-01
issue-536

abder:


Naaaaiix (changes requested):
Keep '.maxi-blocks' as a general class for the MaxiBlocks blocks on the same place they will find on the frontend. That way we keep that consistency 👍

Naaaaiix (approved):

PR #553, opened 2020-10-05
Issue 548

abder:
Fix the issue mentioned in #548 

**Note.**

- I have made the gap smaller, I have noticed that I have been using double value. 
- Also I have changed the vertical gap from 1em to 0.7em to make it as close as possible to the horizontal one.


Naaaaiix (approved):
Good job men, excellent 👍

Naaaaiix (approved):
Gorgeous, spectacular, amazing and incredible! Are enough adjectives? Congratulations men, is a pleasure to read this code 👍

PR #563, opened 2020-10-08
Issue 484

javadsamiee:
Add upload image icon over Image Block placeholder


abder (approved):
Good Job!  

I think in future we may need to add the possibility to insert an image from a URL.

Thinking of something that looks similar to the native Gutenberg Image Block.

Screenshot from 2020-10-08 17-10-37

Naaaaiix (approved):
Good job 👍 

We should come back for this, IMO is not much intuitive and users will find problems.

PR #564, opened 2020-10-08
Issue 532

abder:
This PR aims to close( [#527, #526, #532, #554] ) :smile: 

Please test everything with all possible scenarios and different videos and links from Youtube & Vimeo. 

Thanks :pray:


javadsamiee (commented):
Good job 👍 

- I wonder why you removed the video upload, so if we want to add a video from the local how should we do it? it is not related to you, I think it has a bad UX

- Also, this part has not good style see others input styles

image

- How do I can control 'Autoplay'? after insert video, it starts to play and I can not stop it

Naaaaiix (approved):

PR #566, opened 2020-10-12
Issue 485

javadsamiee:
some minor fix


Naaaaiix (approved):

Naaaaiix (commented):
The solution is good, but has a lack of consistency IMO. Removing the 'setBackgroundStyles' method and replace it for 'getColorBackgroundObject' generates a inconsistency in two ways:
1. BackgroundDisplayer component is useless.
2. Block is using the background in another way, different one. 
These two points generate a future problem: if we want to add more options to backgroundControl, like image or video or svg, we'll need to redo this part. So, I think this is an unnecessary inconsistency easy to fix. That's why I've generated this alternative branch with a possible solution that, IMO, puts more eficency and generates a more solid base; check it out [here](https://github.com/yeahcan/maxi-blocks/commit/865ed855d4bd79cd81f6a49e11f5056da908a785) 👍

Naaaaiix (approved):

PR #573, opened 2020-10-14
Issue-374

abder:
This PR aims to close #374 

**Summary** 
Fixed style issues in RTL mode, we should keep an eye on that mode while developing new components and blocks, a good workflow is that everyone should test on RTL after working on any issue :smile:


Naaaaiix (approved):
Code looks good, but didn't test it in action. Please, @javadsamiee, can you? You've got more experience on it 👍

javadsamiee (approved):

PR #576, opened 2020-10-16
Issue-571

abder:
This PR aims to close #571 

Note: 
- As the native `text-control` component doesn't provide a way to validate user inputs, I created our own `text-control` that we can use instead, right now it supports 'video-url' and we can extend it to support other types as needed. 

- I am not expert in Regular Expressions :smile: if you have any suggestions to improve, please share them


Naaaaiix (approved):
Good job 👍  not 100% about the TextControl structure, but it works for the moment, so let's develop it as it will be necessary and adapt to the new necessities

PR #582, opened 2020-10-19
Issue 579

javadsamiee:
Some styles fix

### DIFF
--- a/src/blocks/column-maxi/editor.scss
+++ b/src/blocks/column-maxi/editor.scss
@@ -1,3 +1,12 @@
+.maxi-column-size-control {
+	.components-base-control__field {
+		flex-direction: column;
+		align-items: flex-start !important;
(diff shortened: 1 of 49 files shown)


Naaaaiix (approved):
Good job 😄

Naaaaiix (commented):
The function is working, but I'm not sure about what the code concern. Too much changes of names for the same info ('SVGCurrentElement' and 'SVGCurrentShape' are the same), and 2 different way to used doesn't make it much consistent (the object that the component 'SVGDefaultsDisplayer' is returning is treated different on Image Maxi and on BackgroundControl). This things work, but at the end generate confusion for readers. 

If you consider is right, let's merge it 👍

Naaaaiix (approved):
Oh, much better! Good job buddy 👍

PR #588, opened 2020-10-20
Issue 587

javadsamiee:
Some fixes


Naaaaiix (approved):
Found SVG option for BackgroundControl on Container Maxi hover; may we should remove the SVG option in all hover situations because is not going to work 👍 

The rest is really good. Strange to see the toolbar without arrow 😸

PR #591, opened 2020-10-26
Font Icon Control Component 

abder:
This pull request introduces a new component called: **font icon control** which is a part of the issue #558, so I wanted to share the solution I made for the icon picker, I appreciate your feedback and suggestions. 

The component with this version supports font-awesome library and we can extend it easily to support more font libraries in the future. 

## The Main Problem with FA
Me and @elzadj have been looking for a way to load the last version of FA all the time without having to update the plugin. 
so we needed to two elements: 
- The font files 
- The list of icons


Naaaaiix (approved):
Reviewed and pushed a version with fixed bugs I found 👍

javadsamiee (approved):

PR #594, opened 2020-10-27
New motion effect

javadsamiee:
It just has implemented for the Maxi Image block, after pass the reviewing and testing level will add it to other blocks.


Naaaaiix (commented):
Magnific piece of art, amazing! love it ❤️ 
Some small comments and suggestions, but the work is huge and awesome. I really congrat you men!

On frontend and UX aspects:
1. The element gets the position on the backend; not sure if it can be confusing. Maybe we should provide some kind of preview feature, but this way is kind of strange:
motioneffectux
2. Disable 'load preset' if there are no presets for loading because is empty
Captura de pantalla 2020-10-30 a las 10 56 32
3. I don't understand what this options are doing, I miss some information. Also, the X, Y and Z, are px or %? What is the measure? In this case, if possible, I would recommend %.
Captura de pantalla 2020-10-30 a las 10 56 32

Naaaaiix (commented):
Good, MotionControl looks much cleaner and really easy to read. About preview: I don't understand what I should expect with it. When I put, it just moves the image:
motionpreview
Also, IMO, measures should be on '%', not in 'px' or 'em': proportion is easier to generate responsive solutions. Also, I don't know what I'm doing; I miss information about the commands and what they are doing:
Captura de pantalla 2020-11-02 a las 10 48 29

Naaaaiix (commented):
When generating an Image Maxi block returns me an error. Checked on incognito also 👍

Naaaaiix (commented):
C'mon, last details! Really good job men!

PR #607, opened 2020-10-30
Issue 603

javadsamiee:
Some fixes

### DIFF
--- a/src/blocks/image-maxi/attributes.js
+++ b/src/blocks/image-maxi/attributes.js
@@ -129,6 +129,9 @@ const attributes = {
 		type: 'string',
 		default: '',
 	},
+	clipPathCurentElement: {
+		type: 'number',
+	},
 	motion: {
 		type: 'string',
 		default: JSON.stringify(attributesData.__experimentalMotion),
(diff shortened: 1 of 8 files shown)


abder (approved):
Looks much better :+1:

Naaaaiix (changes requested):
Looks amazing, but found some things to improve:
1. The 'aria-pressed' is based on a new attribute you just put on Image Maxi, not in BackgroundControl, for example. I see different points: setting a new attribute while you can do the same with comparing, can generate issues if we add more shapes because the number is different, and all of this is more coding and actions. Generate an alternative solution, have a look [here](https://github.com/yeahcan/maxi-blocks/tree/c88f8930300af266e7e845f63c479aa08ac4db29).
2. Modifying the VisualEditor canvas and putting margins and so on, generated a lot of problems with the drag and drop:
clippathrestyled
3. The buttons are showing bad shapes, as circle: is cutted on the top:
Captura de pantalla 2020-11-02 a las 10 15 55

Naaaaiix (approved):
Fantastic! Looks amazing! 😄

PR #612, opened 2020-11-03
issue-599

abder:
This PR aims to close #599


javadsamiee (commented):
I think you should not just disable these things, it's better to have others in the advances tab or show them via something such as more settings under the component

javadsamiee (approved):
Good job 👍

Naaaaiix (approved):

PR #616, opened 2020-11-04
Issue 606

Naaaaiix:
This PR aims to fix #606 . Some considerations:
1. This implementation is for the layers section. With it we'll have the chance to keep improving BackgroundControl, inasmuch as, IMO, BackgroundControl with this layer feature provides a confusing user workflow and, due to it, a confusing UX. This implementation of the layers is a way to see the big potential of the component, and from here we can generate a better controller.
2. This implementation doesn't work with Hover yet! Let's see how this evolves and implement it when is 100% decided how it will work.

Some improvements derived from this implementation that will be generated as issues:
1. Opacity on BackgroundControl doesn't work correctly with hovering; when hovering on the block, the opacity dissappears. Also, the selector for 'yes/no' on the hover section doesn't have anything selected until you click:
backgroundopacity
2. Image Maxi Hover Animation is still rendered wrong:
imagehovererror
3. We start to need a reestructure (no refactoring) of the plugin and the data. Files like the 'edit.js' on the blocks may be divided with the 'getObject' and subsequence methods into other files, and some of the elements inside of the folder 'extension' may need to be reorganized, like the file 'utils.js' on the folder 'style' that is getting too much big. Other things we should consider are related with the 'Styling API' and some other backend process.
4. Videos doesn't fit correctly either on backend and frontend. IMO we should avoid iframes; @DOUARA did a hard job with it, but I consider is not perfect (here's mr. Perfectionist, sorry). There are some methods to fix it with the code we've got, but we'll still depend on iframes and with JS for frontend, something that we should avoid IMO. So, with a fast check I've found [this](https://github.com/u-wave/react-vimeo) dependency that allows Vimeo videos without iframe and with the possibility to set and 'end-time', and we may look for something related for youtube. Here some examples:
- First rendering with youtube video shows controls:
youtubebackground
- Youtube errors on console(*):
youtubevideoerror
- Wrong iframe height for a youtube video on backend:
Captura de pantalla 2020-11-04 a las 11 10 35
- Youtube frontend height error and console errors(*):
yotubefrontendsizeandconsole
- Vimeo backend; orange line shows is not reaching the top:
Captura de pantalla 2020-11-04 a las 11 17 20
- Vimeo on frontend doesn't fit on width:
vimeofrontend

(*): errors on console has been tested on localhost and learndd to compare different situations and enviorments.


abder (changes requested):
I am really impressed with that :heart: this component has become so powerful we should name it `background-builder` instead :smile: 

Here're my thoughts  

> We start to need a reestructure (no refactoring) of the plugin and the data. Files like the 'edit.js' on the blocks may be divided with the 'getObject' and subsequence methods into other files, and some of the elements inside of the folder 'extension' may need to be reorganized, like the file 'utils.js' on the folder 'style' that is getting too much big. Other things we should consider are related with the 'Styling API' and some other backend process.

I agree with the fact that any function in `extensions/styles/utils.js` file that's related to only one block, should be moved to the block folder in `utils.js` file for example. we can reduce the size of that file this way. 
as most functions are `component-related`, we may create a folder in `extensions/styles` named `utils` and forEach function we create a folder for it with the name of the component that'is related to. this way is gonna be easy to maintain and understand 


> Videos doesn't fit correctly either on backend and frontend. IMO we should avoid iframes; @DOUARA did a hard job with it, but I consider is not perfect (here's mr. Perfectionist, sorry). There are some methods to fix it with the code we've got, but we'll still depend on iframes and with JS for frontend, something that we should avoid IMO. So, with a fast check I've found this dependency that allows Vimeo videos without iframe and with the possibility to set and 'end-time', and we may look for something related for youtube.

- IMO it's not possible to make the video fits the height all the time without relying on JavaScript at least, we have to change the Iframe height depending on the width as we need to keep the ratio 16:9 (if the user is using 16:9 video). if not I am not sure if there's a way to get the video aspect ratio, but most of the time it is 16:9.
- For avoiding using Iframes, I don't think there's another way than that as it's related to the service provider, [that library](https://github.com/u-wave/react-vimeo) is actually using iframes under the hood: 
Screenshot from 2020-11-04 15-52-45

Even if it's not, I think we can't use it either, it's a React component and we can only use it on the backend, we need to generate raw HTML on the `save` function for the front-end, then for Vimeo for instance we need front-end JavaScript to interact with it. 

I think the only thing we can do is to find a better way to structure our front-end JavaScript, right now we have all the code in one big file: `js/front-scripts.js`, my suggestion is that for every block we would have a file named: `front-scripts.js` then with webpack we compile all these files in one file inside `dist` folder.  of course editing the Webapck file is gonna be a challenge. 


> First rendering with youtube video shows controls. 
I am not sure if it's possible to hide those meta information, I searched for a way to remove them but I didn't find any. also other plugins like EL have them.


> Youtube errors on console(*):

Those errors are coming from the adblock extesion in my testing: 
Screenshot from 2020-11-04 16-04-18

>  Wrong iframe height for a youtube video 

I think that can be fixed 

> - Vimeo on frontend doesn't fit on width, Vimeo backend; orange line shows is not reaching the top

I am not sure if we can make this perfect, other plugins have the same issues: 

Screenshot from 2020-11-04 15-43-32


## My Notes.
- I think this should be hidden when use layer value is yes: 
Screenshot from 2020-11-04 14-58-38

- Found this bug when setting `custom` size for the background image
captured (3)

javadsamiee (approved):
Good job man 👍

PR #619, opened 2020-11-05
Issue 614

javadsamiee:


Naaaaiix (approved):
Perfect! The only thing I would say is related with the front-scripts.js file: needs some refactoring and enhancements to ensure a perfect UX. But what I'm finding on it are the same problems that other functions on the file are having also, so let's do all work when the file will be 100% fulfilled and ready.

Good job Mr. Motions 😄

PR #682, opened 2020-11-12
Issue 656

javadsamiee:
Add border hover status


Naaaaiix (approved):
Good job buddy! 👍 

Want to share with you a thought: what you've done is a good implementation for this issue, but I'm not sure if we may need to think, when we'll be doing the refactoring in future, that may we need to generate in all the objects that we are using for the components (border, background, boxShadow, etc...) a `status` item as you've done, to give them a coherence. Is related and part of the process of improving the speed and the performance on the Style API.

PR #689, opened 2020-11-16
Issue 665

javadsamiee:
Some fixes


Naaaaiix (commented):
You've generated different elements like:
- `<div className='toolbar-item__column-pattern__popover'>`
- `<div className='toolbar-item__column-size__popover'>`
- `<div className='toolbar-item__divider-alignment__popover'>`
that are attending on the same place and having the same responsibility: a minimum width of '330px'. Shouldn't be better, instead of giving to each element this styling, to have an element inside of the component ToolbarPopover that will get this responsibility and set it just once instead of repeating same style code in different places? Not sure, maybe is better as it is: just a comment/suggestion 👍

PR #699, opened 2020-11-18
Issue 667

javadsamiee:
Some fixes


Naaaaiix (commented):
Didn't check the code, found some bug on frontend:
defaultDisplay
The attribute 'defaultDisplay' on the component 'displayControl' is for this situations. Instead of displaying a 'display: block !important', the component should set a 'displa:flex' in some way. Why? Because if not, it breaks. Would happen the same with Row, Column, Container and others.

elzadj (approved):

Naaaaiix (approved):
Oh yeeeeaaaah! Really good men, finally we've got it!

Can you please merge master on it and test it before we merge this branch into master? thanks 😄

PR #702, opened 2020-11-19
Issue 272

javadsamiee:
- Components from class to function

- Remove Advanced Range Control component

- Fix Shape Divider and Hover Effects background problem

- Some minor fixes


Naaaaiix (approved):
Goooood, a little bit of refactoring! hahahahaha I love it!

PR #723, opened 2020-11-25
Issue 646

javadsamiee:
Some fixes


Naaaaiix (changes requested):
Can you check it again? it breaks on my computer:
cropimagebgerror

PR #724, opened 2020-11-30
Issue 712

javadsamiee:


Naaaaiix (approved):
The implementation is good, but has some issues on performance and lifecycle. As it works, I'm not going to request changes, and let's leave it as it is: we'll fix it in future 👍 Anyway, it will be readapted with #727 and with the future DB and API works. Good job!

Naaaaiix (commented):
Bye

javadsamiee (commented):
Magnificent and clean organization 👍

elzadj (changes requested):
Bye

elzadj (changes requested):
Bye

elzadj (approved):
Let's merge it. We always can catch and fix small bugs (if there are any) later.

PR #737, opened 2020-12-03
Issue 734

javadsamiee:
I have updated Wiki as well, but we should update it regularly and add documentation for components that have not it, as I suggested before we should do it every week


Naaaaiix (changes requested):
Good job! Please, have a look to uppercases and camelcases depending the situation 👍

elzadj (approved):

Naaaaiix (approved):

PR #740, opened 2020-12-04
Issue 739

javadsamiee:
Add units for line size
Some design improvement
Fix resize problem
Add tooltip for resizing and show the height of the divider


Naaaaiix (approved):
Amazing implementation 😃 I love the new tooltip, and the way you structured the 'edit.js' file abstracting the functions out of the render part and with the state.

Would like to know why didn't set units on weight also:
Captura de pantalla 2020-12-09 a las 10 24 25

PR #744, opened 2020-12-08
Issue 724

javadsamiee:
I'm not sure about my solution but I tried to fix defaults and responsive in editor responsive changes and frontend view


elzadj (changes requested):
Tested - and the custom font size does not work, it gets stuck on the default value, because the CSS rule with var(--maxi-light-h1-font-size-s) !important  rewrites the custom font.

Desktop-screenshot

Also a side note: IE 11 does not support vars in CSS - are we going to use a polyfill?

elzadj (approved):

PR #746, opened 2020-12-08
Issue 743

javadsamiee:


elzadj (commented):
Great work!

Naaaaiix (changes requested):
Good job! Agreed with @elzadj: on clicking the button the modal should open. It will increase the UX 👍

elzadj (approved):
Thanks! Much better. We have a lot of small bugs to fix, and need to re-write it to use the Cloud API in the future, + inheriting the settings when we change the icon,  but it's not urgent.

PR #747, opened 2020-12-09
Issue 736

javadsamiee:
Add Highlights


elzadj (changes requested):
Cool, I have some notes:

1. I think we need one custom setting for the highlight, with a multiply selector:
--->Highlight: text, bg, border<--- 

Or we can move settings @javadsamiee created around - add highlight text to typography, border - to border, bg - to bg. But I think the first way with just one setting is more usable.
This can wait for now tho. Not important for the library. 

2. We may want to make the text and bg highlights to disable each other. I.e. if the text is on - the bg is automatically off, and vice versa. Again, this can wait.

3. There is a problem with the border - when you highlight it, nothing really happens, because it's None in the settings. Need to make it solid automatically, I think. This we need for the library. @javadsamiee , please, add it.

4. I do not see the highlight classes on the frontend. I think we need to test Frontend too, at least for one block, to make sure it's ok before we start the patterns.

elzadj (changes requested):
Need to update the border settings, not just add another border. The highlight toggles the border colour only.

Naaaaiix (changes requested):
Container and Row return me an error 💥

Naaaaiix (approved):
Cool! Did some suggestions for improving, but is working, so it can be merged if my suggestions are not considered 👍 😄

Naaaaiix (changes requested):
On testing Container it breaks 👍

Naaaaiix (approved):
Looks good for me =) The only thing would be what @elzadj commented: when setting a highlight bg, may the text should become other colour. Would you like to merge this branch and set a new issue for it or what's the plan?

elzadj (approved):
Merge it, we can deal with the text later :+1:

PR #764, opened 2020-12-16
Issue 759

javadsamiee:
Some fixes


Naaaaiix (changes requested):
`src/components/font-level-control/index.js` has references for margin on the state and on props. Also there are some component instances with the attribute 'marginOptions' 👍

elzadj (approved):

Naaaaiix (approved):

PR #766, opened 2020-12-16
Issue 745

elzadj:
#745 
Default container for fullHD is 1170px now.
For 4k - 1790px
Changes in columns gaps to match the psds. 
Fix for Row and Container width on backend.
Fix for column gaps on mobile.
Full width is disabled by default.


Naaaaiix (commented):
I have my doubts about this implementation:
1. Row gets values for 'max-width', but is not editable.
2. Row, without a full-width option set, is covering full-width on my 13" screen. That makes the theme invalid, and is incoherent with the Gutenberg flow; this can be confusing for users.
3. What to do if I want a Row smaller?
4. What would be the workflow for a user generating a pattern with this configuration?
5. Setting the width of 'sizeContainer' in all responsive stages makes Container loses its power of being a double container. The inner container is the same size than the outer container. With it, how can we generate a fullwidth background with content inside?

Not sure if this configuration will cover all our necessities 👍

Naaaaiix (changes requested):
When loading a Container Maxi it returns an error 👍

javadsamiee (approved):

Naaaaiix (commented):
Some issues:
1. On changing column sizes, is not keeping the value after saving:
Kapture 2021-01-14 at 10 22 57

2. Now that Container Maxi integrates directly the Row, maybe we need to reestructure Container Maxi. For example, I'm wondering if the elements '.maxi-container-block__wrapper' or 'maxi-container-block__container' are still necessary or they just disturb.
3. Image doesn't fit with the ResizableBox:
Captura de pantalla 2021-01-14 a las 10 35 51
Captura de pantalla 2021-01-14 a las 10 37 42
4. On set 'FullWidth' to Row Maxi, it just moves to left side and doesn't fit all the Container:
Captura de pantalla 2021-01-14 a las 10 45 05
5. How can I change the Row size? I mean: there are no options for 'sizeContainer' on Row, so it means users can just deal with our default sizes or have a full-width. Is it like this?

Naaaaiix (approved):
Cool, minor things that once fixed make the branch ready to be merge.

I'm wondering: @elzadj and @javadsamiee , are you using eslint and prettier on your VSCode? I see sometimes lots of tab 'errors' that this two plugins are fixing automatically in my local. Is a lot of time that I'm saving and make me work much faster; ping me if is not working for you or you have doubts or suggestions about it 👍

PR #798, opened 2020-12-23
Issue 793

javadsamiee:
some fixes


elzadj (approved):

Naaaaiix (changes requested):
Not sure if this solution is fixing the issues; look what I'm finding:
1. On setting an image layer, the image appears twice.
2. If you generated a layer on other block, the same options are copied to the next one.
Kapture 2020-12-28 at 11 13 03

Naaaaiix (changes requested):
We are soooo close!

Naaaaiix (approved):
Cool! 😸

PR #800, opened 2020-12-23
Issue 796

javadsamiee:
Minor fix


elzadj (approved):

Naaaaiix (commented):
Don't understand what has been done. When are we going to use the advanced options related with 'disableAdvanced' prop?
My suggestion on #796 is that, when border-style is 'none', the rest of options doesn't appear:
Captura de pantalla 2020-12-28 a las 11 37 14

Naaaaiix (changes requested):
Cool, works good! But found this:
Kapture 2021-01-05 at 09 52 28

Naaaaiix (approved):
Oh yeah! That's right! 😄

PR #801, opened 2020-12-24
Issue 794

javadsamiee:
Minor fix


elzadj (changes requested):
Remove the console.log pls

Naaaaiix (changes requested):

elzadj (changes requested):
Please test the column's width on frontend - it gets a wrong value.

elzadj (changes requested):
I still see that bug with the columns' width on the frontend.

PR #812, opened 2021-01-04
Issue 811

javadsamiee:
Minor fix


elzadj (changes requested):
Cool! Can you please fix it for the column's bg too? I think that's it, a Container and a Column only. Rows are ok.

elzadj (approved):

PR #813, opened 2021-01-04
Issue 807

javadsamiee:
Minor fix


elzadj (changes requested):
I think we can add it to the editor.scss only? We don't need it on the frontend.

elzadj (approved):

PR #832, opened 2021-01-20
Optimisation for Container and Divider

javadsamiee:


Naaaaiix (commented):
Really good implementation, just few details. In editor works fine, just one thing: highlights doesn't work, the components needs to be update. 😄

Naaaaiix (approved):
Cool, let's merge into the optimization branch!

PR #837, opened 2021-01-21
Issue 787

elzadj:
- Add Remove Gap Setting #787 

Most changes are in /src/components/column-pattern/index.js
- Change margin for responsive columns from 0.7em to 1.5em. We can change it later to whatever we want, but 0.7 is too small.


Naaaaiix (approved):
Looks good for me, I think is a really good implementation, I like it! Good job 😄

PR #843, opened 2021-01-22
Issue 841

javadsamiee:
Some fixes


elzadj (approved):
Working much better now, thanks! Don't forget to add the same max-width to the normal Row, please.

Naaaaiix (approved):
Cool! If @elzadj says is ok, for me is perfect (in terms of values). The code has a comment, but not a big deal as it works 👍 Good job!

PR #846, opened 2021-01-26
Issue 836 optimized

elzadj:
Group Block implementation #836


Naaaaiix (approved):
First look it seems nice to me, the only thing I would say is that it could be confusing for users as we will have Container and Group that will do basically the same. But you know better than me 😄

javadsamiee (approved):
Good job Svit 👍🏻

PR #842, opened 2021-01-22
Optimization for Button Block

javadsamiee:

PR #847, opened 2021-01-26
Optimization for Text Maxi

Naaaaiix:
Alignment, highlights and some buttons on the Toolbar are not active because we should merge #842 where has been done and implement here; the rest should work good 👍


elzadj (approved):
Works nice, just need to resolve the conflicts :+1:

javadsamiee (commented):
Good job buddy 👍🏻 

Under typography, Size doesn't work, and text-shadow when I change x,y, blur doesn't have any effect also typography hover don't work.

javadsamiee (approved):
Superb 👍🏻

PR #849, opened 2021-01-28
Image Maxi, Text Maxi and Toolbar optimization

Naaaaiix:
Sorry, couldn't test as much as I would like. Anyway, the only known error is with Divider direction Toolbar component 👍 

FYI: this branch will merge into 'javad-optimisation-suggestion' branch, because it has been merged into it to use some optimized components and, in case we would merge into master will introduce 'javad-optimisation-suggestion' changes into master. So, it can't be merged into master until 'javad-optimisation-suggestion' will be perfect, this merged into it, and after 'javad-optimisation-suggestion' to master 👍


javadsamiee (commented):
Good job man, 👍🏻  I have found some minor issues 😄 

- [x]  Background doesn't work
- [x]  Width in Toolbar image Size doesn't work
- [x]  Image size in sidebar -> after the reset with its value field is empty, is it correct behavior? 
- [x]  Edit image button inside Toolbar image Size have not correct target to the sidebar settings, it has been targeting width/height
- [x]   Under custom image caption slider control of Size & Line Height don't work
- [x]  Margin Top, I can not check the Auto option 
- [x]  Box Shadow doesn't work in the sidebar and toolbar
- [x] https://github.com/yeahcan/maxi-blocks/pull/849#issuecomment-769623579
- [ ] https://github.com/yeahcan/maxi-blocks/pull/849#issuecomment-769622770

PR #851, opened 2021-02-01
Javad optimization image frontend

javadsamiee:


Naaaaiix (approved):
Found some errors on changing the image size. I would suggest to stop struggling with it during the optimization process and fix it when optimization will be done, because image size is related to #827 and #840 . Here's and example of the error I'm finding:
Kapture 2021-02-01 at 11 55 43

The rest works simply fine, thanks for fixing all this errors buddy 🙏

PR #857, opened 2021-02-03
Issue 853

javadsamiee:


Naaaaiix (changes requested):
Not going to keep reviewing cause I see lots of necessary actions:
1. Clean correctly the methods on the folder 'src/extensions/styles' => stylesResolver.js is not used anymore, same for stylesGenerator.js; they have been upgraded to newStyleResolver.js and newStylesGenerator.
2. Update the file 'src/extensions/styles/index.js' for a good re-exporting.
3. On files of 'attributes.js', remove 'newAttributes' label.

👍

Naaaaiix (approved):
Cool, couple of minor things! let's fix them and move on 😄

PR #862, opened 2021-02-03
Image cop control fix

Naaaaiix:
Fixes [this commented error](https://github.com/yeahcan/maxi-blocks/pull/842#issuecomment-771462937)


javadsamiee (commented):
Good job man, it's working but is it normal that the image width is 0 in the initial?

image

and when I click over the image without creating a shape to crop I will face this one

image

We should have a way to cancel crop, here it is not possible and we will face with this error

PR #863, opened 2021-02-03
Image min width for XXL and S

elzadj:
#840


javadsamiee (approved):
Good job 👍🏻

Naaaaiix (changes requested):
I would suggest to use another type of implementation that avoids duplicating 'size' attributes. Look how it has been done in [Button Maxi 'attributes.js'](https://github.com/yeahcan/maxi-blocks/blob/e201f3f18cf2df3400003a9abdaa2e11a08fa9ba/src/blocks/button-maxi/attributes.js#L29-L58) for example.

PR #866, opened 2021-02-03
Fix Column size delay

Naaaaiix:
Found an error when changing Column Maxi size:

https://user-images.githubusercontent.com/50477222/106787193-b3355b00-664f-11eb-93e9-880f5c221192.mp4

So, I've been checking the [ResizableBox](https://github.com/WordPress/gutenberg/tree/004a32ec09d5b789be694d85725fe49cca385f08/packages/components/src/resizable-box) again, and as I knew it was a copy of other component called [Re-Resizable](https://github.com/bokuweb/re-resizable#props) from where we have been getting the documentation. Tested before to use the `ref` of the 'ResizableBox' in order to use the API that allows to modify the resizable element from any place. That would make us avoid using imperative orders (`document.getElement...(...).style.width = '...'`) and use state programming as React requests. But I couldn't use it before, and now that I saw the error that was affecting this precisely I thought it was the moment. So, installed 'Re-Resizable' into our plugin and tested on Column Maxi. It works really good IMO, and is much faster than the older method (the older method was requesting every change for the element/node to change using `document.querySelector(`.maxi-column-block__resizer__${uniqueID}`).style.width = `${val}%`;`).

So, couple of things:
1. As we are not using Gutenberg component 'ResizableBox', we lost some classes that makes the CSS of this new component doesn't be affected. @javadsamiee can you use your magic to help me with it? You'll find how Gutenberg made this component in [this page](https://github.com/WordPress/gutenberg/blob/004a32ec09d5b789be694d85725fe49cca385f08/packages/components/src/resizable-box/index.js)
2. This same process can be reproduced in blocks like Divider Maxi that has effects like this:

https://user-images.githubusercontent.com/50477222/106788629-7c604480-6651-11eb-9647-a7538c0e4288.mp4

Another good point is that we are not going to depend from Gutenberg versions for this component, so their changes won't affect us on future 👍


elzadj (changes requested):
I'm getting 
react_devtools_backend.js:2430 TypeError: Object(...) is not a function
    at edit.render (edit.js?3a05:111)
    
which is bizarre :thinking: 

Screenshot from 2021-02-04 13-58-33

javadsamiee (commented):
The reset button act weird and show number 90 after reset

PR #879, opened 2021-02-04
Issue 874

javadsamiee:


elzadj (changes requested):
Great work!

elzadj (approved):

Naaaaiix (changes requested):
Well, not sure about this implementation. Some points that we should review:
1. Removing the file 'src/extensions/defaults/typography.js' generates problems when pasting content to Text Maxi because is necessary. Have a look [here](https://github.com/yeahcan/maxi-blocks/blob/0efd0afca978e52e6ed44eb1602f4350b67a04fa/src/blocks/text-maxi/edit.js#L15)
2. Derived from 1: we need a global default file with all the defaults for typography.
3. FontLevelControl needs to apply the defaults over the old values, but it needs to keep the last settings in case the user comes back to the old font level. Let's see: I have a Text Maxi with 'p', and put color and font-size. Change to 'h2' and I don't like, so come back to 'p': I would like to find the changes I left before. That's why FontLevelControl used states 👍 

Let's have a look and find a fix that keeps that functionalities 😄

PR #884, opened 2021-02-05
Fix #875

Naaaaiix:
Never seen this before, and may it helps you in future. HTML elements under `position: fixed` can be affected by parent elements with transform properties. In this case, the blocks were under the TranslateControl statement that were setting `transform: translateX(0%) translateY(0%)` and this was affecting to the position of the element. Hectic, first time I see it in my life, and I don't really understand the documentation:

> For elements whose layout is governed by the CSS box model, any value other than none for the transform property results in the creation of a stacking context. Implementations must paint the layer it creates, within its parent stacking context, at the same stacking order that would be used if it were a positioned element with z-index: 0. If an element with a transform is positioned, the z-index property applies as described in [CSS2], except that auto is treated as 0 since a new stacking context is always created.
> 
> For elements whose layout is governed by the CSS box model, any value other than none for the transform property also causes the element to establish a containing block for all descendants. Its padding box will be used to layout for all of its absolute-position descendants, fixed-position descendants, and descendant fixed background attachments.

[more info](https://www.w3.org/TR/css-transforms-1/#example-02d1d5ba)

So fixed TransformControl also 👍  (may related to #859 )

### DIFF
--- a/src/blocks/divider-maxi/edit.js
+++ b/src/blocks/divider-maxi/edit.js
@@ -35,6 +35,7 @@ class edit extends MaxiBlock {
 	state = {
 		isResizing: false,
 	};
+
 	get getStylesObject() {
 		return getStyles(this.props.attributes);
 	}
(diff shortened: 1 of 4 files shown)


elzadj (approved):
Bye

javadsamiee (approved):
Awesome 👍🏻

PR #889, opened 2021-02-08
Issue 880

javadsamiee:
Fix TextShadow Color and Opacity

### DIFF
--- a/src/components/text-shadow-control/index.js
+++ b/src/components/text-shadow-control/index.js
@@ -56,11 +56,10 @@ const TextShadow = props => {
 		else onChange(valueDecomposed.join(' '));
 	};
 
-	const getActiveItem = prop => {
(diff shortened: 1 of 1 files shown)


elzadj (changes requested):
Niiice, thank you.


Can we keep it across all types please?

elzadj (changes requested):
Much, much better

elzadj (changes requested):
Better, but I still see these

Naaaaiix (approved):
Works perfect for me 😄

elzadj (changes requested):
I still see that bug with the previous colour applying :thinking: 

@Naaaaiix @javadsamiee can you re-create, or is it just me?

elzadj (approved):
Wooooorks.

PR #902, opened 2021-02-09
Issue #858

Naaaaiix:
Fixes #858


javadsamiee (commented):
I can see the error yet and also display don't work in the toolbar as well.

elzadj (changes requested):
> I can see the error yet and also display don't work in the toolbar as well.

I confirm both bugs.

javadsamiee (approved):

elzadj (approved):

PR #907, opened 2021-02-10
Issue 859

elzadj:
https://github.com/yeahcan/maxi-blocks/issues/859#issuecomment-773934768

Closes #859


Naaaaiix (approved):
Good job 👍 😄 

The only thing I would suggest, but can be derived to #906 is that, on the helper, if the attribute is equal to it's default, don't set it. I mean, if a rotate value for X is 0, is not necessary to set it. Is for avoiding CSS rules like:
Captura de pantalla 2021-02-11 a las 10 33 48
That doesn't do anything 👍

javadsamiee (approved):
Great 👍🏻

PR #915, opened 2021-02-12
Issue 913

javadsamiee:
Fix for rotate and AxisControl


Naaaaiix (commented):
Onces I set a value, I can set 0 but not empty 👍

Naaaaiix (approved):
MAGNIFIC! 😄

elzadj (approved):

PR #919, opened 2021-02-15
Issue 831

javadsamiee:
Control min/max for input number


Naaaaiix (commented):
mmmmhhh... not sure if it's a good idea this implementation. If we have to do this every time we've got to use RangeControl will be really tedious. What do you think if you generate and abstracted component that does this job automatically and change all the instances of the original RangeControl to this new abstraction? I think it will help and make developing faster than being integrating this code every time we use RangeControl 👍

Naaaaiix (approved):
Looks good for me, much better! Thanks @javadsamiee 😄

elzadj (approved):

PR #928, opened 2021-02-15
Fix #900

Naaaaiix:
This branch has 2 implementations:
- Fixes #900 (not a priority, but took me 2 min)
- Create higher order component for FormatValue blocks: the way we were obtaining the `FormatValue` object was creating errors on console and was rendering much times Text Maxi. First step for fixing other issues of Text Maxi 👍

### DIFF
--- a/src/blocks/text-maxi/edit.js
+++ b/src/blocks/text-maxi/edit.js
@@ -122,8 +122,6 @@ class edit extends MaxiBlock {
 			className
 		);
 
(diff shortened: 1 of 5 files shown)


javadsamiee (commented):
Good job man 👍🏻 

RTL is not correct and the RTL mode text-align still is LRT

image

elzadj (approved):

javadsamiee (approved):
Superb 👍🏻

PR #933, opened 2021-02-16
Text Maxi fixes

Naaaaiix:
1. Fixes #927 
2. Fixes #925 
3. Fixes errors on pasting content

There's a limit on selecting that comes directly from RichText and happens the same on the native Paragraph block. On selecting part of text of one block and part of other, it selects all text of both blocks. I'm sure is fixable, but not easy 👍

### DIFF
--- a/src/blocks/column-maxi/inspector.js
+++ b/src/blocks/column-maxi/inspector.js
@@ -4,7 +4,7 @@
 const { __ } = wp.i18n;
 const { InspectorControls } = wp.blockEditor;
 const { Fragment } = wp.element;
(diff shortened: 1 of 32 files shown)


elzadj (changes requested):
Great :+1: 


react-dom.min.js:15 Uncaught TypeError: Cannot read property 'max' of undefined
But maybe it's a different issue? 

- Also Ctrl+A does not work in the Text Maxi, and other text-selecting keys like "Shift + arrow". They start working if you add a basic paragraph block to the page.

- Clear the console logs please :smiley_cat:

javadsamiee (approved):

elzadj (approved):

PR #948, opened 2021-02-17
Image Maxi size suggestion

Naaaaiix:
Not sure if it fixes #827 , maybe a part. Anyway, basic idea is:
1. ResizableBlock works with '%', not with 'px'. So the range bar on the sidebar settings panel, inside of ImageDimension, works correct now. 
2. Width/Height affect the <figure> element, not the image. So the sizes are different. This is the same that we have for all blocks: width/height affects the wrapper element of the block. Image size (point 1) affects the element  (well, not really: affects the hover-wrapper <div>, but is like the same)

I would suggest to remove the sizes selector (thumbnail, small, medium, full...). But I would like to know your opinions about before keep ~~destroying~~ implementing it 👍


javadsamiee (commented):
Good job 👍🏻 

Image size works in sidebar good but in the toolbar have problem 

image

javadsamiee (approved):
Superb 👍🏻

elzadj (approved):

PR #957, opened 2021-02-18
Issue 921

javadsamiee:
Fix motions effects


Naaaaiix (approved):
The error has been solved, but we have a problem:
- Just 'px' unit works.
- Just works on published posts.

As we talked by videocall, let's derivate these points into new issues and fix them apart. Good job buddy 👍

elzadj (approved):

PR #962, opened 2021-02-18
Issue 954

javadsamiee:
Add selected for columns and remove some extra styles


Naaaaiix (approved):
The idea and implementation is good, but not sure if we should apply to all blocks, not just in columns 🤔

elzadj (approved):

PR #963, opened 2021-02-18
Issue 929 style cards

elzadj:
#929 

What does not work / not ready:

- Reset custom settings
- Style card (SC) editor / importer / exporter
- SC Font Family: 'Roboto' set as default for now.

What works:

- SC Maxi (Light) by default, the Custom setting was removed. Left the 'Parent' for now for the inner blocks.
- Any custom style re-writes the default SC style.
- Highlight re-writes the custom settings.
- Added text color and bg color for Buttons.
- Fixed link / button SC styles conflict (both are using the 'a' tag) 
- Removed doubled SC settings for the link

I think that's enough to try to make our first patterns from the Default (Maxi) - Light psds.


javadsamiee (approved):
Great 👍🏻

Naaaaiix (approved):
I don't know what I like more, its power or its simplicity! Congrats, I think we are accomplishing something really hard: do simple the complex things 💪 🤩 

I have one 'but', but is not a big 'but'. How to deal with transparent background if we have a colour coming from the SC and we can just put other things on it? We should think in a new option inside Background? 🤔

PR #967, opened 2021-02-19
Issue 965

javadsamiee:
Minor fix

I have checked the unit and preview problems those are working correctly, please check it on a new page and make sure for caching.

### DIFF
--- a/src/components/motion-control/showTimeline.js
+++ b/src/components/motion-control/showTimeline.js
@@ -105,23 +105,24 @@ const ShowTimeline = props => {
 								<div className='maxi-motion-control__timeline__group'>
 									{value.map((item, i) => (
 										<div
-											key={`maxi-motion-control__timeline__group__item-${i}`}
+											key={`maxi-motion-control__timeline__group__item-${key}-${i}`}
 											className={classnames(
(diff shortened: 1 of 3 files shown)


elzadj (changes requested):
Motion effect Move - works

Scale - nope, I've also got this error when I've tried to remove it from an image block:
TypeError: Cannot read property '0' of undefined
    at MotionPreview (webpack-internal:///638:26)
    at je (react-dom.min.js:78)
    at qe (react-dom.min.js:89)
    at ph (react-dom.min.js:216)
    at lh (react-dom.min.js:126)
    at O (react-dom.min.js:121)
    at ze (react-dom.min.js:118)
    at react-dom.min.js:53
    at unstable_runWithPriority (react.min.js:26)
    at Ma (react-dom.min.js:52)
    
Rotate - nope (I've tried z-axis)
Skew - nope, same error as with Scale

elzadj (changes requested):
Checked Skew - it still is not working for me. Can you please record a video of how you make each of the Motion Effects work on an Image? Thanks.

elzadj (approved):

Naaaaiix (approved):

PR #974, opened 2021-02-22
Issue 968

javadsamiee:
Add FullSize option on the toolbar


Naaaaiix (approved):
Perfect 🥳

elzadj (changes requested):
Nice work!
Looks like the Font Icon and the Group Blocks are missing the setting (both in the toolbar and the sidebar).
Also the Image align does not work, do we have an issue for that?

elzadj (approved):

PR #987, opened 2021-02-23
Fixes #976

Naaaaiix:
Fixes #976

### DIFF
--- a/src/components/background-control/backgroundLayersControl.js
+++ b/src/components/background-control/backgroundLayersControl.js
@@ -212,6 +212,11 @@ const BackgroundLayersControl = ({
 							onDragEnd={(fromIndex, toIndex) => {
 								const layer = layers.splice(fromIndex, 1)[0];
 								layers.splice(toIndex, 0, layer);
+
+								layers.forEach((layer, i) => {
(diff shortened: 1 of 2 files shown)


elzadj (changes requested):
Great! Just one thing - the layers are in the wrong places on backend (frontend is ok):

Screenshot from 2021-02-23 13-12-23
Screenshot from 2021-02-23 13-12-48

elzadj (changes requested):
I think the layers are added in the reverse order. I've added Color - Image - Shape - Video, but they are shown as Video - Shape - Image - Color. 

If it was made intentionally, it's really confusing - you add a solid color, you add an image, but you do not see the image, because the previously added solid color is on top. To see the image, you need to manually move it.

Look at the screenshots:

Screenshot from 2021-02-23 13-56-31
Screenshot from 2021-02-23 13-56-09
Screenshot from 2021-02-23 13-54-51
Screenshot from 2021-02-23 13-54-34

Also the block validation fails after I add a layer with a YT video - maybe need a new issue about it :+1:

javadsamiee (approved):

elzadj (approved):

PR #996, opened 2021-02-23
Divider Maxi suggestion

Naaaaiix:
Made some changes on Divider Maxi for suggest some implementations that make it be more coherent with the rest of the blocks. Having a width of '50px' all the block generates situations like:
Captura de pantalla 2021-02-23 a las 17 25 07

This branch propose to make it look like:
Captura de pantalla 2021-02-23 a las 17 26 01

### DIFF
--- a/src/blocks/divider-maxi/attributes.js
+++ b/src/blocks/divider-maxi/attributes.js
@@ -38,12 +38,6 @@ const attributes = {
 	},
 	...{
 		...attributesData.size,
-		'width-general': {
-			default: 50,
-		},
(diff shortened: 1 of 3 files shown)


javadsamiee (approved):
Good idea 👍🏻

elzadj (approved):
We just need it to look as close to our psds as possible to make life of the editors easy. Also a divider is always going to be inside a Container.

Alejandro9kgm (approved):
the bug has been fixed but I have found another one, now it generates a new issue.

PR #1022, opened 2021-02-26
Issue 1021

javadsamiee:
Fix Reset buttons
Change border-radius to value base
Hide line settings in None mode


Alejandro9kgm (approved):

Naaaaiix (commented):
Some detail:
I don't understand why the change for the line border radius from being a button to be a range slider; not sure if it's a good idea for the UX. Anyway, it doesn't work for me: the CSS is there, but can't see it in the result.

PR #1078, opened 2021-03-05
Issue 1055

javadsamiee:
Add NumberInputControl to fix cross-browser input type number problem, also fix for uncontrolled inputs such as empty and 'auto' 

@Naaaaiix Please review the codes I just put it for z-index in the sidebar if everything was ok I can use it for other places
image


Naaaaiix (approved):
Cool, good job!
I saw some think I wanted to suggest you, but it was easier to generate an alternative branch. Have a look at it [here](https://github.com/yeahcan/maxi-blocks/tree/issue-1055-alternative). Some of the changes:
1. The use of the 'control' on the components is used normally for the ones that have a 'base-control', normally have a label and inputs. Some examples are 'SelectorControl' or 'RangeControl'. This new input components doesn't fill with the requirements.
2. Some options like placeholder, ariaLabel, roundBy or allowAuto has been added.
3. Set the type on the buttons and `type='number'" for the input.

Anyway, tested on Chrome, Safari, Firefox and Edge and looks amazing! Good job buddy 😄 

And as a comment:
the use of 'useEffect' to change states is kind of tricky, but it helps with the uncontrolled/controlled inputs, so good choice!

PR #1101, opened 2021-03-09
Issue 1095

javadsamiee:
Minor fix


Alejandro9kgm (approved):
now you can remove the hover animation, but these buttons don't work. 
Captura

PR #1129, opened 2021-03-11
Clear IE and remove extra dependency

javadsamiee:


Naaaaiix (approved):
back on time! lol Good one, goobye ie 🤣

PR #1230, opened 2021-03-29
Implement Jest

Naaaaiix:
In order to start testing automatically, Jest has been implement on our environment. Is using the native WordPress 'e2e-test' dependency for Gutenberg, so we don't need to config anything.
Some important aspects:
- In order to generate test we can use [React Test Renderer](https://reactjs.org/docs/test-renderer.html) or [Enzyme](https://enzymejs.github.io/enzyme/). Please, refer to its documentation and [Jest documentation](https://jestjs.io/docs/getting-started) for more info 👍 
- As we are under CGB boilerplate, is really hard to config anything. In this case, config Jest to work with renamed dependencies is not possible, so we can't use the imports based on global variables as we made until now. In practice:
```jsx
// Not supported by Jest
const { RangeControl } = wp.components;

// Supported by Jest
import { RangeControl } from '@wordpress/components';
```
- Added 2 new scripts to npm: `npm test` and `npm test:debug`. First runs testing, second opens a new Chrome console to debug.

Also, set a couple of random tests on `ColorControl`. You'll find lots of examples for tests on Gutenberg package 😄


javadsamiee (approved):
Another super fantastic work from Josh 😄 🚀

Alejandro9kgm (approved):

PR #1240, opened 2021-03-31
Custom template

javadsamiee:
image

I have tested it with Divi, Astra, OceanWp, GeneratePress, and Kadence and this concept worked good, please test it and let me know your points


Alejandro9kgm (approved):

elzadj (approved):
Good job! Now we have a solution for new pages.

Naaaaiix (approved):
Good idea @javadsamiee ; next thing could be advice users when they are using other templates and tell them that if they don't use our template they can experience differences between the Cloud Library demo designs and the designs they'll find on their frontend 👍

PR #1272, opened 2021-04-19
Issue 1262

javadsamiee:
Add replace icon for Font Icon


Naaaaiix (changes requested):
Good job @javadsamiee , just found that the button appears even when I'm not hovering 👍

Alejandro9kgm (approved):

PR #1284, opened 2021-04-20
Issue 1259

javadsamiee:
Change font selector

### DIFF
--- a/src/components/font-family-selector/editor.scss
+++ b/src/components/font-family-selector/editor.scss
@@ -1,56 +1,14 @@
 .maxi-font-family-selector {
 	display: block;
-	&__button {
-		height: 31px;
-		border-radius: 5px;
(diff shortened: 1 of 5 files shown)


Naaaaiix (commented):
Really love the job you've done: this component was something I was wondering to refactor and you made a fantastic job. Congrats!

Couple of things:

- [x] On hover, there's no cursor effect. May a `cursor: pointer` would make it more accessible.
- [x] The searcher is strange: doesn't invite you to use it. As you can see in the video, the caret appears, but as the current font is on the background, is hard to understand you can search:

https://user-images.githubusercontent.com/50477222/115417321-d1dfd280-a1f8-11eb-8268-7b613136e087.mp4

Alejandro9kgm (approved):

PR #1286, opened 2021-04-21
Basic tests e2e

Alejandro9kgm:
These are the first tests for each Maxi Block, they are simple tests to know if the blocks break


Naaaaiix (approved):
Good job! May we need to accurate a little more in future, but let's see how they comport and after decide to improve if necessary 👍

javadsamiee (approved):

PR #1294, opened 2021-04-23
Issue 1265

javadsamiee:


Naaaaiix (commented):
Cool, good idea and good implementation. Not merging because it's not compatible with the work I've been doing for Text Maxi and needs to be adapted, so as soon as Text Maxi optimization will be released I'll adapt this code to make it run with the block 👍

Alejandro9kgm (approved):

Alejandro9kgm (approved):

PR #1305, opened 2021-04-26
Set empty input when search on FontFamilySelector

Naaaaiix:
When open the selector of Font Family it turns the input on empty in order to allow users understand they can search for inside 👍


javadsamiee (approved):
Good solution man, thx 👍🏻

PR #1326, opened 2021-04-28
Issue 1313

javadsamiee:


Alejandro9kgm (approved):

Naaaaiix (approved):
Not a bad implementation, but there's a fear that surges on me. `hoverPreview` component is made to fit `Image Maxi`, and looks it won't be reusable with other blocks or components. Not a big deal now our plugin is 'small' and we know what we do, but I'm afraid we'll need to refactor it if we ever need to reuse this component in another place 👍 

For the rest, as always, good job buddy!

PR #1334, opened 2021-04-29
Issue 1306

javadsamiee:
Minor fix

### DIFF
--- a/src/blocks/button-maxi/edit.js
+++ b/src/blocks/button-maxi/edit.js
@@ -75,8 +75,13 @@ class edit extends MaxiBlock {
 			'maxi-block',
 			'maxi-block--backend',
 			'maxi-button-block',
-			getLastBreakpointAttribute('display', deviceType, attributes) ===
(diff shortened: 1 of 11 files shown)


Alejandro9kgm (approved):

Naaaaiix (commented):
It may fix the issue, but I'm not sure about the implementation. On `getLastBreakpointAttributeGroup`, in the mapping of the line 72, you're iterating on the element `attributes` of the first element selected, not in the attributes of every selected item. That make `getLastBreakpointAttributeGroup` same than `getLastBreakpointAttributeSingle`, so doesn't have any sense.

The main objective of `getLastBreakpointAttributeGroup` is to return a value in case all the blocks share same result. If not, returns null. We need to be really carefully when changing this functions, because they are  the core and we can fix one thing but broke 100 👍

Naaaaiix (approved):

Alejandro9kgm (approved):

PR #1348, opened 2021-05-03
Issue 1337

javadsamiee:
1- Change parallax
2- Remove GSAP and from this point Motions won't work


Naaaaiix (commented):
Cool! I have some doubts:
1. Is normal that in backend doesn't show or preview anything?
2. Is normal that in frontend the image, on moving with the parallax, leaves a white background?

Alejandro9kgm (approved):

PR #1371, opened 2021-05-05
Adapt Maxi for all WP versions

Naaaaiix:
IMPORTANT
This branch is not going to be merge until SC are merged. There are too much changes and can generate incompatibilities.

This branch aims to make Maxi Blocks compatible with all the existent versions of WP since 5.5.3 . So it works for:
- 5.5.3 (probably 5.5, 5.5.1 and 5.5.2 also)
- 5.6
- 5.6.1
- 5.7
- 5.7.1

The good part of this implementation is that we don't need to maintain different plugins for every version, so everything is in one. Some of the optimizations done:
1. Generate a space on the API and selector on the `maxiBlocks` store to access global settings of Maxi. In this case, returns the Gutenberg version. In future we'll be able to detect if users are also working with the `Gutenberg` plugin and not just with the one installed on WP.
2. Replace the way we were accessing to `anchorRef` on Toolbar using a `ref` and fixed much problems with `ref` on blocks.
3. Abstract `InnerBlocks` in a new component to deal with the different versions of Gutenberg and adapt all blocks that uses this component with the new one.
4. Fix and improved the actions related with the DOM, like removing the native Gutenberg toolbar.

What is not done:
1. Didn't fix the styles.
2. Didn't review every deep detail.

So, to test:
- EVERYTHING lol

How to do it easily: (I recommend this way)
0. With docker on your computer
1. In console: `npm run wp-env start` (will download WP 5.5.3)
2. In console: `npm start`
3. Test that version
4. To change version: change it on `/.wp-env`
5. Repeat from step 1

So, hopefully you'll like as much as me. So happy to see Maxi on WP 5.7.1 😸


javadsamiee (commented):
Maxi Icon should not be here, It should be left of the header

Pasted_Image_5_6_21__1_17_PM

javadsamiee (approved):

Alejandro9kgm (approved):

Alejandro9kgm (approved):

javadsamiee (approved):

PR #1372, opened 2021-05-06
Issue 1349

javadsamiee:


Alejandro9kgm (approved):

Naaaaiix (approved):
I would define your job as beautiful: really good code Javad. Clean, direct and effective; both parts: backend and frontend. Congrats!

In terms of frontend there are some things:
1. When adding the block, would look amazing that it moves until 100 slower. It does the effect but really fast 😄 
2. Duration range input didn't work until I've pressed reset button. On changing it doesn't seem to change the speed or the duration.
3. Every time I change something, I have to press the button to reproduce. It may be better to reproduce automatically, don't you think?
4. In frontend the animation is too fast, can't see it.
5. Radius may should be limited to the size of the container.
6. May would be nice to be able to select different fonts 👍

Alejandro9kgm (commented):
For some reason it does not exceed 1%


https://user-images.githubusercontent.com/58038966/117613591-6c339600-b167-11eb-95c1-4469118cd32b.mp4

[code.txt](https://github.com/yeahcan/maxi-blocks/files/6449877/code.txt)

Naaaaiix (commented):
I like this block more every day, is stunning! 😸 

Couple of suggestions:
1. When adding a block, is it possible to auto-reproduce the first time?
2. Should be disabled to set end < start
3. Font size can be bigger than the circle, not a problem, but strange.

BTW: amazing job with the radius and stroke!

Naaaaiix (approved):
Last thing... sorry 😅 

Duration is appearing in miliseconds, not sure if it's what expected 👍

PR #1376, opened 2021-05-06
Issue 1373

javadsamiee:
Minor fix


Naaaaiix (commented):
What do you think if instead of that we set something like:
`
    margin-left: auto;
    margin-right: auto;
    width: 100%;
`
That way we ensure that the top&bottom are set by the default WP CSS and we are not interfering on it 👍 🤔

Alejandro9kgm (approved):

Naaaaiix (approved):

PR #1377, opened 2021-05-06
Issue 1368

javadsamiee:
Minor fix

### DIFF
--- a/src/blocks/column-maxi/style.scss
+++ b/src/blocks/column-maxi/style.scss
@@ -4,9 +4,4 @@ body.maxi-blocks--active div.maxi-column-block {
 	flex-direction: column;
 	height: auto;
 	margin: 0;
-
-	& > *:not(.maxi-background-displayer) {
-		position: relative;
-		z-index: 1;
-	}
 }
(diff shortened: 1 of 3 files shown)


Naaaaiix (commented):
Not sure about this implementation. We are putting `z-index` everywhere and every time we put one we have a new issue. Let me explain a little:
This error we are fixing is related to `src/blocks/column-maxi/style.scss:20` where we put a `z-index: 1` to fix the Paragraph inside the Column on #1321. So, when removing this z-index we fix the issue we are treating in this PR. And I would say that #1376 would fix it also. So, instead of keeping fixing the errors we are generating fixing other errors, I would suggest next implementation:
Related with #1321, #1376 and this PR: instead of setting `z-index: 1` on `src/blocks/column-maxi/style.scss:20`, use [this suggested comment](https://github.com/yeahcan/maxi-blocks/pull/1376#pullrequestreview-653481164) to fix it and #1376 . This method should fix all 3 problems without keep playing with z-index 👍

Alejandro9kgm (approved):

PR #1379, opened 2021-05-06
BaseControl && SelectControl

Naaaaiix:
This merge has 2 implementations related with #1375 :
- BaseControl
- SelectControl

What to test:
- BaseControl: all components with names like `{component}-control`. Radio buttons, inputs, selects, range inputs, etc...
- SelectControl: selectors in general.

Please, test them on different WP versions. Remember this branch is not related with #1371 , so has errors when working with newer versions of WP. All the content has been copied from Gutenberg 8.5, the version WP 5.5.3 is using 👍


Alejandro9kgm (approved):
In the toolbar the size item, its texts are in gray and not in white

Captura de pantalla 2021-05-07 082651

javadsamiee (approved):
Great work, with this solution we will have not any problems in future 😄 🎉

PR #1382, opened 2021-05-07
Fix #1381

Naaaaiix:
Fixes #1381


javadsamiee (approved):
I got this error after remove block has 
`
componentDidMount() { this.interval = setInterval(() => { this.setState({counter: this.state.counter + 1}); }, 1000); }
`

image

PR #1451, opened 2021-05-18
Issue 1449

javadsamiee:


Alejandro9kgm (commented):
It feels very weird to hide the circle, and see how the rounded bar option disappears and the hide circle options moves up, I think it would be nice to change the order and put the hide circle option first.

Captura de pantalla 2021-05-18 092638

I also find it strange that the rounded bar option disappears but not the Circle Background Colour and Circle Bar Colour options.

https://user-images.githubusercontent.com/58038966/118610387-00cb7300-b7bc-11eb-98ea-c1541834ecf0.mp4

Alejandro9kgm (approved):

Naaaaiix (approved):

PR #1468, opened 2021-05-19
Keep global values on SC

elzadj:
Closes #1425


javadsamiee (approved):

Alejandro9kgm (approved):

Naaaaiix (commented):
Instead of introducing a new entry to SC object with the `${props}-old` and make it bigger, with more functions and conditionals, wouldn't be better to use the existing `${props}-global` to decide if the property `prop` should be displayed or no? So, keep the value on `prop`, and let `${prop}-global` as point to render or not render `prop`.

What I see as a consequence of this implementation is a bigger SC object, more conditionals into the functions and the components, that translates into a more complex structure to maintain, to debug, to fix, to scale and to update. 👍

PR #1474, opened 2021-05-20
Style card editor affects Gutenberg blocks

elzadj:
Quick fix to exclude other blocks from our SC styles. 

Closes #1426

### DIFF
--- a/src/css/editor.scss
+++ b/src/css/editor.scss
@@ -84,6 +84,14 @@ body {
 			box-shadow: none !important;
 		}
 	}
+	&:hover {
+		.block-list-appender {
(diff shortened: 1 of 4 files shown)


Naaaaiix (approved):

Naaaaiix (approved):

Alejandro9kgm (approved):

javadsamiee (approved):
I should say I dont like this file, very long and not optimized, this file in the future will be fatter, so it is better to optimize and classify it now. for example, we can use separate files and imported them into maxi-themes, and also add comments to understand it better.

When I wanted to add this file I should think and find a correct place

PR #1552, opened 2021-06-01
Issue 1516

javadsamiee:
Add warning box component


Naaaaiix (approved):
The only suggestion I would do is that, maybe, instead of `WarningBox` I would call the new component `InfoBox`, as is not a `warning` like: ERROOOOOOR! FIIIIREEEE! lol is more like: `hey, you know? just consider it` 😆 

But if you don't consider we merge it as it is 👍

Naaaaiix (approved):

Alejandro9kgm (approved):

PR #1581, opened 2021-06-07
Issue 1472

Naaaaiix:
This PR aims to fix #1472 and #1571 

Important notes:
1. SC object has been updated in 2 directions: as it has been talked on #1472 to let other blocks and parts have space there and to be easier to work with, and in Typography object way, to avoid parsing the SC object to work with TypographyControl in both directions.
2. `maxiStyleCardsTab` has been refactored to be able to manage point number 1.

Known bugs:
!IMPORTANT! => SC doesn't work on frontend: the reason is easy. When adopted the note number 1 and applied Typography obj on SC object, is necessary to parse it. For this process has been used the same helper parser we use for Style API (`src/extensions/styles/helpers/getTypographyStyles.js`), but in this case it just works on backend as it's part of JS. Relating it to #1542 , but more with #1352 , the future workflow won't process the Style API and the SC API on frontend, because is a waste of time and resources that will affect really negative on SEO. The idea is to save the SC styles as a `meta` string of the object as we do with the Style API and just get it when some visitor request the post on frontend. Would be something really similar to the way WP renders the HTML. So, in order to don't advance that process and make this PR insanely big, I left it to do onces this is approved and merged or at least reviewed and approved, to be sure we are agree on this implementation 👍 
=> SC doesn't renders correctly until the SC panel is opened: related to the first point, as is not rendered on frontend, doesn't show SC styles until the SC panel is opened 👍 

PD: is necessary to clean SC from DB to make it work correctly 👍


javadsamiee (approved):

elzadj (changes requested):
1. Looks like the switching between text levels' font size does not work until I open the SC Editor (see the video):

https://user-images.githubusercontent.com/2401618/121026949-3f4ac100-c7af-11eb-8506-d8649b5f4981.mp4


2. The default button is broken after I reload the editor

```
Uncaught TypeError: Reduce of empty array with no initial value
    at Array.reduce (<anonymous>)
    at getWinBreakpoint (index.js:437)
    at showNotification (index.js:454)
    at index.js:530
    at Array.map (<anonymous>)
    at index.js:513
    at renderWithHooks (react-dom.js?ver=16.9.0:15246)
    at mountIndeterminateComponent (react-dom.js?ver=16.9.0:17480)
    at beginWork$1 (react-dom.js?ver=16.9.0:18624)
    at HTMLUnknownElement.callCallback (react-dom.js?ver=16.9.0:341)```

elzadj (approved):

Alejandro9kgm (approved):

PR #1600, opened 2021-06-09
Issue 1577

javadsamiee:
I removed NumberControl & RangeSliderControl and replaced them with SizeControl, in the future we can add more settings like type and hide some parts of them, also with this implementation we have fixed remove numbers inside of input number and we have an Integrated input number


Naaaaiix (approved):
Cool! 👍 The only thing I would say is related to the component label: SizeControl. It born as a component to control the sizes, and in some occasions we disable the unit, but not common behavior. Now is easier to have it with the unit disable than to have it with. So, is not much a SizeControl anymore lol IMO we may think on refactor it and change the name in some moment and avoid future confusions 👍

Alejandro9kgm (approved):

PR #1601, opened 2021-06-09
Issue 1543

Naaaaiix:
Fixes #1543 

Please, @elzadj , have a look because there's some reason it was like that and I broke something with this implementation 😄


javadsamiee (approved):

elzadj (approved):
Looks like it does not break anything

PR #1602, opened 2021-06-09
Cloud API - Algolia

elzadj:
Adds Algolia search to our Cloud Modals.


javadsamiee (commented):
Much better!

Naaaaiix (approved):
Much better!

Alejandro9kgm (approved):

Naaaaiix (approved):
What a pleasure to have the Cloud Maxi working; amazing job! 😸 

Left some comments, but any is important. We can derivate them into smaller issues and merge this branch with no problem 👍 

Couple of comments for all the devs:
1. After seeing this branch's PHP structure (for example, when opening a function the bracket is in a new line), it shows we need to look for a way to standardize it in for all the team. If not, it appears as a change and makes harder the reviews. Let's generate an issue for it.
2. About the image saver implementation and its class, there's something it worries me: dirty the creators WordPress. Is good the implementation considers not repeating images, but I would suggest to think if it's possible to set to all that images a `taxonomy`  to be easy for the users find all Maxi images to deal with them as they want.

Alejandro9kgm (approved):

javadsamiee (approved):

PR #1696, opened 2021-06-17
Cloud API - placeholder image, taxonomy

elzadj:
Closes #1675

Note: only this one has a valid Gutenberg code (at the time of the creating this PR) for tests:

Edit-Page-‹-Local-Wordpress-—-WordPress


Naaaaiix (approved):
Well done, really clean and easy to read 😸 

Couple of notes:
1. When requesting placeholder images, it returns an error. I don't have the same problem with the normal images, so I guess is not related with my directory permisions:

https://user-images.githubusercontent.com/50477222/122523374-3d86c600-d017-11eb-8816-2df8aa1dfe94.mp4

2. As a suggestion and with the idea to follow the rest of the plugin structure, I would suggest to move `uploader.js` and `placeholder.js` into a unit file with a name `util.js` or move them into a new folder called `library` on  `extensions`. The main idea is to difference the function and React components when we have this folder. With the files as they are right now, seems like `uploader.js` or `placeholder.js` are components, not functions 👍

Naaaaiix (commented):
The filters on backend looks good, make much easier the way to find and do bulk actions on Maxi Images. I wonder if it's possible to put it in another way, as creating Maxi Images as a hole taxonomy generates the possibility to set new taxonomies in this kind of box that could generate confusion:
Captura de pantalla 2021-06-21 a las 10 56 24

Naaaaiix (approved):
Good job 😸

javadsamiee (approved):

PR #1722, opened 2021-06-23
Issue 1693

Naaaaiix:
Finally has been a massive refactoring, sorry for the problems this will cause in your working branches. At least, the implementation has entered to master with no conflicts, so it means it shouldn't be a big pain in the ass (luckily!).

Well, this implementation does:
- Allow palette colors in responsive stage
- Fix some incoherences related with palettes
- Fix some bugs (mostly related with the helpers)
- Update tests

Some comments:
1. Kept the palettes colors as classNames in `general` responsive stages, in the rest is pure Style API that generates CSS classes. For example, if creators change the font color in `l` responsive stage, the CSS rule is `color: --var(...)`. We may think if is better to remove that `general` classNames, as they consume lots of bits (much more than CSS variables) and require more time for maintenance  and are also another piece that can break easily in case of bug (less functions, less errors).
2. There are some classes on `palette.js` and `getPalettesClasses.js` that I profit to remove, as they were related with `icon`. Also found some instances of `icon` on Button Maxi.
3. Created new objects for `svg`, `map` and `numberCounter`.
4. IMPORTANT => Not fixed the part related with SVGs on BackgroundControl as it needs to be fixed also in other issues like #1684


javadsamiee (approved):
Outstanding 👍🏻

elzadj (changes requested):
Niiice :smiley_cat: 

Some bugs I've found:

- [x] Cannot change background back from a custom color to palette. Tested on buttons and text. Border, shadow and text color work ok, so only backgrounds are affected.
Video: 

https://user-images.githubusercontent.com/2401618/123630582-603a8b00-d81e-11eb-8434-71a4e25597fe.mp4

- [x] Can we set the default custom color to the currently used palette color? Or something else, I'm not sure. You can see the problem in the video above (time 0.03) , when I switch the button's bg to custom color, it shows write bg until I choose the custom blue color. 
- [x] Style Card Editor crashes when I try to apply a color (I've cleared my SCs in DB):

```
Uncaught TypeError: colorString.charAt is not a function
    at getRGB (index.js:56)
    at ColorControl (index.js:75)
    at renderWithHooks (react-dom.js?ver=16.13.1:14938)
    at updateFunctionComponent (react-dom.js?ver=16.13.1:17169)
    at beginWork (react-dom.js?ver=16.13.1:18745)
    at HTMLUnknownElement.callCallback (react-dom.js?ver=16.13.1:182)
    at Object.invokeGuardedCallbackDev (react-dom.js?ver=16.13.1:231)
    at invokeGuardedCallback (react-dom.js?ver=16.13.1:286)
    at beginWork$1 (react-dom.js?ver=16.13.1:23338)
    at performUnitOfWork (react-dom.js?ver=16.13.1:22289)
```



Video: 

https://user-images.githubusercontent.com/2401618/123631647-ab08d280-d81f-11eb-85f5-7a3309d55228.mp4


- [ ] Maxi light hover is #c9340a, but it's not set by default. Makes border's hover on buttons white.

elzadj (approved):
>  a new issue

Yes to all, please add them :smiley_cat: 
Great work!

PR #1728, opened 2021-06-25
Issue 1555

javadsamiee:
Show "Add Row Maxi" next to the "Add Block" and change its style to rounded, please check this solution, the problem goes here is when we have not any row the "Add Row Maxi" button is not centered if I have a class when container have Row I can handle this one also

If you like this solution we can find a good way to put it center


Naaaaiix (approved):
Not a bad idea... maybe there are other ways, but for the moment helps a lot. Good job 😸

PR #1740, opened 2021-06-29
Issue 1682

Naaaaiix:
Fixes #1682 and #1265 . Some considerations:
1. Default color palette for 'active' and 'visited' has been set randomly as there's no indication on [this spreadsheet](https://docs.google.com/spreadsheets/d/18yHSXVRHlggTw_fVFu82kyNDFcCO9kMoyj1BtkcOzUA/edit#gid=0)
2. Just set on Text Maxi, but should be able to adapt to any other block without a problem


elzadj (changes requested):

elzadj (changes requested):
Not sure why we do not have the default color 5 for h1-h6 - now they always stay gray (well, color 3, same as 'p'). 
--maxi-light-h6-color-palette: var(--maxi-light-color-3);

Maybe we need a new issue for adding the default color numbers and checking them?

elzadj (approved):

Alejandro9kgm (approved):

javadsamiee (approved):
Good job 👍🏻

PR #1743, opened 2021-06-30
Issue 1736

javadsamiee:


Naaaaiix (commented):
Well, the implementation works, but IMO is too much process and too much conditionals for something that could be much simpler. It comes to my head a function inside 'src/extensions/style-cards' with a name `getSCPropValue` that returns the value of the current SC, with parameters like `target, blockStyle, SCentry`. Onces this function returns the value for the globals, we can send it as a prop for `ColorControl` and this blurs the palette if necessary. This way we don't need to do a big round requiring that amount of information on `paletteControl.js` and that amount of conditionals. I'm thinking something like:
```
// inspector.js
// blockStyle = light
// textLevel = 'p'
<ColorControl 
...
globalStatus={getSCPropValue('color-global', blockStyle, textLevel)}
/> 
```
```
// src/extensions/style-cards/getSCPropValue.js
const getSCPropValue = (target, blockStyle, SCEntry) => {
    const selectedSC = select('maxiBlocks/style-cards').receiveMaxiSelectedStyleCard().value || {};

    return selectedSC.?[blockStyle].?styleCard[SCEntry][target] || false;
}
```
```
// paletteContro.js
[...]
	const paletteClasses = classnames(
		'maxi-sc-color-palette',
                globalStatus && 'palette-disabled'
        );
```

Maybe I'm wrong and I'm doing a mistake, but really think the process suggested is much faster, easier and will generate less problems in future. Imagine the same with more options on SC and more blocks: it will increase the conditionals and the necessary information exponentially !

Share your thoughts, as I said, I might be wrong 🤔

Alejandro9kgm (approved):

Naaaaiix (approved):
Much better and simpler, thanks @javadsamiee 😄

PR #1763, opened 2021-07-05
Remove SVGDefaultsDisplayer

javadsamiee:


Alejandro9kgm (approved):

Naaaaiix (commented):
Some bugs I've found:

1. BoxShadow doesn't work correctly with Image Maxi + shape:
Captura de pantalla 2021-07-06 a las 16 58 40
2. SVG background layer, on fill option, doesn't work (I guess that's why on `e2e-tests/components/background-control/__snapshots__/index.spec.js.snap:56` the "background-svg-SVGData" appears empty).

PR #1793, opened 2021-07-16
Issue 1776

Naaaaiix:
Fixes #1776 

Create some new cases for Text Maxi and its Custom Formats API:
1. When having 2 Text Maxi, each one with its Custom Formats, the classes can coincide, and onces merged creators can lose their styles. That has been fixed on this branch.
2. When having a whole content link on Text Maxi, if creators decide to remove the link from a part of the content, they lose the link and the styles get crazy (underline keeps, etc...). That has been fixed on this branch.

All cases are accompanied by their respective tests 😄


Alejandro9kgm (approved):

javadsamiee (approved):
Great tests 👍🏻

PR #1822, opened 2021-07-20
Issue 1818

javadsamiee:
Fix for border radius and border

### DIFF
--- a/src/blocks/column-maxi/styles.js
+++ b/src/blocks/column-maxi/styles.js
@@ -18,25 +18,6 @@ import { isEmpty } from 'lodash';
 
 const getNormalObject = props => {
 	const response = {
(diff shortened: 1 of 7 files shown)


Naaaaiix (commented):
Some comments:
1. All styles set to elements like `' .block-editor-block-list__layout'` are useless on frontend, so we should avoid referring them on the Style API as we are generating unnecessary CSS code we are loading on frontend.
2. The result is not working as expected:
Captura de pantalla 2021-07-20 a las 16 38 17
3. Didn't we removed the `'borderRadius'` from the background helper? 🤔 

We need a solid solution for this issue as is one of the basis of the plugin and needs to be easy to adapt to future blocks 👍

PR #1828, opened 2021-07-21
Bg + Border possible solution

Naaaaiix:
Here's a possible solution for the big issue we're carrying with the Background and the Border styles. Just done on Group Maxi, and didn't work in it much deeper, as I would like to know your opinion. Is not common, but it works, at least in all the cases I've tested.

This PR is not going to be merged; first need to be validated as a possible solution and then applied to all blocks 👍 

PD: didn't update the tests, probably will break some 👍


elzadj (approved):
The border looks nice and I do not see any issues, at least visual.

javadsamiee (approved):
I like your solution, it's working good, let's apply it for other blocks and see the result for others as well

Alejandro9kgm (approved):

javadsamiee (approved):

elzadj (approved):

Alejandro9kgm (approved):

PR #1835, opened 2021-07-22
Issue 1826 remove link

elzadj:
Closes https://github.com/yeahcan/maxi-blocks/issues/1826


Naaaaiix (commented):
Is not a bad plan to have that conditional, and solves the problem, but can generate others. I recommend to add, as is in the rest of the 'text' related components in the Toolbar, to send to `TextLink` component the prop `styleCard` and from there, add when the component requests for `getCustomFormatValue`. That way, we ensure that this method has the opportunity to compare with SC also, and don't generate unnecessary code 👍 😄

Naaaaiix (approved):

Alejandro9kgm (approved):

javadsamiee (approved):

PR #1865, opened 2021-07-27
Issue 1861

javadsamiee:
Fix border on hover status


Naaaaiix (commented):
Cool, but not sure if this implementation solves what the issue was exposing; probably didn't explain correct myself.
The logics fo the border, at least as I understand, is: if I have a normal border, and I active the hover and set the hover border on `none`, when hovering the element shouldn't appear any hover; right? So that's what the issue was trying to expose. What do you think devs? @elzadj @javadsamiee @Alejandro9kgm

Alejandro9kgm (approved):

Naaaaiix (approved):
Cool, seems good 😄

Alejandro9kgm (approved):

PR #1875, opened 2021-07-28
Issue 1805

Naaaaiix:
Fixes #1805 and updates Text Maxi tests in order to avoid false positives. 

Caution: the speed improvement is amazing, so don't write too fast in Text Maxi or you can break the barrier of the speed of light!!!


javadsamiee (approved):
Super fast 🐇

Alejandro9kgm (approved):

elzadj (approved):
Much, much faster! :+1:

PR #1912, opened 2021-08-02
Issue 1892 - Middle and top margin on responsive

elzadj:
Fixes #1892 

Let's see if 2.5% is better for the margin-top.


Naaaaiix (approved):
Looks good, but as the solution uses `%` on the vertical axis is kind of 'casualty' or more an attempt of approximation (that doesn't mean is bad, in fact is that is much better than what we had before!). Didn't try, but comes to my head another way to play with. Considering the units we have, maybe we can use 'vw' instead of '%' for a more accurate result. Could be a 'complex' calc, but considering that by default the `max-width` of the container is `90%` of the x-axis (equal to `90vw`), we can calculate the percentage of the gap (that affects on horizontal and generates the width of the columns) inside that `90%`. I mean:
```
const unit = '%';
const containerMaxWidth = 90;
const containerReduction = 100 - containerMaxWidth; // 10
const gapBetweenColumns = 2;

const columnsMarginTop =  gapBetweenColumns - (gapBetweenColumns * 10 /  100); // 1.8 (vw)
````

Tested changing the CSS directly and seems it can work 🤔

javadsamiee (approved):

PR #1921, opened 2021-08-03
Issue 1871

Naaaaiix:
Fixes #1871 adding hover default options for bg, border and box-shadow as the normal options 👍


elzadj (changes requested):
I experience some weird behavior: 

- [ ] border-radius does not work
- [ ] Hover does not inherit the non-hover values


https://user-images.githubusercontent.com/2401618/128011469-afe5f6fa-ea20-43d6-a2a4-2fcfdecdb479.mp4

elzadj (approved):

javadsamiee (approved):

PR #1945, opened 2021-08-07
Issue 1929

Naaaaiix:
Fixes #1929


javadsamiee (approved):
Good job 👍🏻  maybe it will be better to show the defined place that moved item will move there with focusing with border highlight or anything

elzadj (approved):

PR #1954, opened 2021-08-09
Issue 1919

javadsamiee:
Fix arrow should have at least a background


Naaaaiix (commented):
Well done! Please, update tests 👍

elzadj (approved):
Nice work @javadsamiee 

The arrow is getting border's width 2x bigger than the container (30px instead of 15px on the screenshots:

Edit-Page-‹-Local-Wordpress-—-WordPress (21)
4k-test-–-Local-WordPress

I see that it was done on purpose, but not sure why - looks weird.

https://github.com/yeahcan/maxi-blocks/blob/7ab853bfc466549a7685ad7fbb39238b788f70f0/src/extensions/styles/helpers/getArrowStyles.js#L50

Any ideas why?

elzadj (approved):

PR #1955, opened 2021-08-10
Issue 1918

javadsamiee:


elzadj (changes requested):
Please bring back "sync all 4' options 
Edit Page ‹ Local Wordpress — WordPress (2)

And remove syncing all 4 when we sync left + right and top + bottom.

elzadj (approved):
Much, much better, tx @javadsamiee 
I think the next step is to allow unique unit for each of 4 values, but let's not go there yet :smile_cat:

Naaaaiix (approved):

PR #1974, opened 2021-08-12
Issue 1965

Naaaaiix:
Fixes #1965


javadsamiee (commented):
Excellent work 👍🏻  I think it's better to have a reset for "Colour Opacity"

javadsamiee (approved):

elzadj (changes requested):
I think we need to hide the palette's opacity when we use a Custom color. 
Edit-Page-‹-Local-Wordpress-—-WordPress (35)

elzadj (changes requested):
Thanks @Naaaaiix 

There is a bug with the Opacity number - first I cannot change it with keyboard, and then it breaks with

```
react-dom.js?ver=16.13.1:82 Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.
```


https://user-images.githubusercontent.com/2401618/129337691-d78f0d8a-fbd4-40e9-904c-a5083d7cadaa.mp4


[blocks.txt](https://github.com/yeahcan/maxi-blocks/files/6981428/blocks.txt)

javadsamiee (approved):

PR #1983, opened 2021-08-13
Issue 1906

javadsamiee:


Naaaaiix (approved):
Fantastic, really cool ! 😸 

Some suggestions, maybe we can derive to other issues:
1. Can't drag less to 10px, I mean, can't set 0.
2. Is not easy to know you can drag to move padding and margin, maybe we can set a cursor, or the lines can appear when hovering or even a tooltip.
3. The line to drag is a bit out of the box:
Captura de pantalla 2021-08-16 a las 9 37 57

Naaaaiix (approved):
Good job Javad! UI is not the best, but we'll fix it in future 👍

Alejandro9kgm (approved):

PR #2006, opened 2021-08-17
Issue 1985

javadsamiee:


Alejandro9kgm (approved):

Naaaaiix (commented):
Being honest, I'm not understanding this fix and its necessity. What is and why we need `maxi-container-arrow__overlap`? With a background color, is not fixing the issue:
Captura de pantalla 2021-08-18 a las 10 52 43

When setting layers it breaks the block also, but not sure if this element `.maxi-container-arrow__overlap` would let the layers be visible also. IMO this way of fixing the issue is overcomplicate and creates non-necessary CSS code. If you consider is correct, please explain it as I couldn't understand it. Thanks 😸 

PD: lots of helpers tests have been removed. That's even intriguing me more 🤔

PR #2007, opened 2021-08-17
Update axis-control test

Alejandro9kgm:


Naaaaiix (commented):
This test is not checking the logics of the component. Please, advise that when you have the `sync all 4`, and then you click on `sync top and bottom` the first one gets unselected, and when you click again `sync all 4` the `sync top and bottom` disappears, but having selected `syn top and bottom` and clicking `sync left and right` both keeps activate. The aim of this test is to ensure all this logic and test all combinations in order to don't break in future. Please, check all the possible combinations and ensure is tested to avoid future problems 👍

javadsamiee (approved):

PR #2010, opened 2021-08-17
Reset SC action

Naaaaiix:
Creates SC reset action on SC store. To test it:
1. Modify SC, then save/apply
2. Refresh and check the changes are kept
3. On console: `wp.data.dispatch('maxiBlocks/style-cards').resetSC()`
4. Refresh and check if SC are default

This feature is just for developing and will be removed with stable 👍


javadsamiee (approved):
Your action works good 👍🏻

PR #2040, opened 2021-08-20
Issue 2037

javadsamiee:


Alejandro9kgm (approved):

Naaaaiix (commented):
The helper `getHoverEffectsBackgroundStyles` is returning styles even when the `Hover effects` is disabled, please ensure it doesn't happen and do a test for it.

PR #2042, opened 2021-08-23
Issue 2015

Naaaaiix:
Fixes #2015
Adapt ResizableBox for SVG Icon block
Implement responsive values for Stroke Width and Width on SVG Icon block
Implement Size settings on SVG Icon Block
Include unit on SVG Width
Replaced `RawHTML` native component for Maxi component on SVG Icon block

Would be good to have a look at this lines and, instead of having that amount of targets on `styles.js` of the block, have them inside of the helper and request it just one time 👍 
https://github.com/yeahcan/maxi-blocks/blob/dcd001da24a796bdac55dc8049818f9d3013417d/src/blocks/svg-icon-maxi/styles.js#L113-L122

### DIFF
--- a/e2e-tests/blocks/column-maxi/__snapshots__/index.spec.js.snap
+++ b/e2e-tests/blocks/column-maxi/__snapshots__/index.spec.js.snap
@@ -1,0 +1,11 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Column Maxi Column Maxi does not break 1`] = `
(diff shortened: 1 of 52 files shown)


javadsamiee (commented):
Good job 👍🏻  very clean and organized

I just say the palette color for SVG line color doesn't work on the toolbar

elzadj (approved):
Looks and works nice, thank you!

javadsamiee (approved):

PR #2061, opened 2021-08-25
Issue 2045

javadsamiee:
Fix Fullwidth shape


Naaaaiix (commented):
Mostly done! That's the big part, so let's see some details:
1. As the `height` of the SVG element on the background layer is `100%`, it takes all the space and generates spaces on the top and on the bottom:
Captura de pantalla 2021-08-25 a las 16 44 53
So, let's set `height: auto` for it.

2. When doing `#1`, we have new situations, like making the size and the position work better. So, let's have a spin on it and make the shape fully functional =)

PD: just checked on Background shapes; be sure it doesn't happen the same on Containers shapes 👍

elzadj (approved):

Alejandro9kgm (approved):

PR #2067, opened 2021-08-27
add responsive test to font-level-control

Alejandro9kgm:


Naaaaiix (commented):
FontLevelControl doesn't have responsive support, so 2 things:
- This tests is not testing FontLevelControl, is testing font-size
- This test is not necessary

PR #2085, opened 2021-09-02
add responsive test to axis-control

Alejandro9kgm:


Naaaaiix (commented):
Cool! Let's see: the test is good, but I would prefer something deeper in this case as this component is really important and really delicate. For the responsive test let's remove the creation of a new post and keep the job was done on the previous test, then jump to 's', do the same process you've done on the normal test (testing different syncs and so on) and check the attributes if they are saved and if the inputs are selected or not. We faced an error of it last week and would be awesome to prevent in future the same problem. Then keep the same than the other tests: check on `m` and check on `xs` the inputs values/selected/etc..

javadsamiee (approved):

Naaaaiix (approved):

Naaaaiix (commented):
Please, clean the code of the util and the branch in general 🙏

Naaaaiix (approved):
Awesome, incredible clean and easy. Congrats, keep working like this 💪

PR #2089, opened 2021-09-02
 Add transition duration control for a button

elzadj:
Adds a hover transition duration control for a button, part of the https://github.com/yeahcan/maxi-blocks/issues/2043 

See the Sidebar - Advanced - Hover Transition.

~~Now, an important question before I proceed with adding the same option for the links inside text blocks~~

~~Do we need to add "Advanced - Hover Transition" to all blocks? We have a lot of blocks that use different bg, border, or text color on hover, but this option will collide with any fancy hover animation we add later. So I'm not really sure we need it aside of the button and links, at least for now.~~

From Kyra: "button and links for now is enough"

Note: we will add Speed Curve option in paid Maxi version. Free one has the duration option only.

### DIFF
--- a/src/blocks/button-maxi/attributes.js
+++ b/src/blocks/button-maxi/attributes.js
@@ -120,6 +120,7 @@ const attributes = {
 	...attributesData.entrance,
 	...attributesData.transform,
 	...attributesData.zIndex,
+	...attributesData.transitionDuration,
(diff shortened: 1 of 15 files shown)


Naaaaiix (approved):
Easy and clean, lovely 😺

PR #2100, opened 2021-09-03
Font weight on frontend

Naaaaiix:
This PR is not linked to any issue as hasn't been create. Fixes the fact that in frontend the `font-weight` wasn't working as expected as the fonts weren't correctly enqueued. That was making, for example, that all the headings didn't show the default `font-weight: 500`. 

This implementation is not optimized. In terms of SEO is horrible as is enqueuing all the font-weight and font-styles, but is working good as for the moment what we need is to prepare the Library. Will be improved when we'll start with the optimization 👍 

How to test?
Test in preview and in published post if the `font-weight` on frontend changes and also with different `font-styles`

### DIFF
--- a/core/class-maxi-api.php
+++ b/core/class-maxi-api.php
@@ -54,6 +54,8 @@ if (!class_exists('MaxiBlocks_API')):
 			$default_array = [
 				'_maxi_blocks_styles' => '',
 				'_maxi_blocks_styles_preview' => '',
+				'_maxi_blocks_fonts' => '',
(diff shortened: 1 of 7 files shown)


elzadj (approved):
Much better, @Naaaaiix !

It sometimes bugs out and I need to re-apply the font (see the video), but works in general:


https://user-images.githubusercontent.com/2401618/132008606-774eae4c-1ae1-435a-8549-d3537d220795.mp4

Alejandro9kgm (approved):

PR #2133, opened 2021-09-13
add responsive test to arrow-control

Alejandro9kgm:


Naaaaiix (commented):
Try to don't create a new post, and use the values of the first test to check the size and position on responsive 👍

PR #2138, opened 2021-09-14
Issue 2102

Naaaaiix:
Fixes #2102 && update tests

The problem with the fonts was that `Roboto` wasn't load when adding a Text Maxi block, so it was necessary to change the font family onces to make it work correctly. Now is fixed 👍 
About the weight, style and the rest of options that didn't work in some situations, was related to a non-expected situation with the maxi custom formats that I have to confess I don't know how it ends like that. Anyway, created a method that solves that situations in case we'll face in future and ensures a better behavior 👍 also created a test to ensure it won't break on future.

### DIFF
--- a/core/class-maxi-styles.php
+++ b/core/class-maxi-styles.php
@@ -122,8 +122,12 @@ class MaxiBlocks_Styles {
 	 */
 
 	public function enqueue_fonts($fonts) {
-		if (empty($fonts)) {
-			return;
+		if (!is_array($fonts)) {
+			$fonts = [];
+		}
+
+		if (!array_key_exists('Roboto', $fonts)) {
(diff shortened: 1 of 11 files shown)


Alejandro9kgm (approved):

elzadj (changes requested):
Thanks @Naaaaiix !

- [ ] When I switch from H to p, the palette color is stuck on the H color (most of the time, but not always): 

https://user-images.githubusercontent.com/2401618/133388711-c03a8df9-2006-4f62-b6df-a03733aff06b.mp4

- [ ] SC Global colors for a link do not work (make a new issue if unrelated, pls)

elzadj (approved):

PR #2170, opened 2021-09-22
Issue 2164

javadsamiee:


Naaaaiix (commented):
Cool! The change is impressive, well done! 🥳 

Some UX situations, like when clicking on `ToggleSwitch` that sometimes doesn't gives the chance to click and you need to accurate and find the exact point for it 👍

kyrapieterse (approved):

Naaaaiix (commented):
Cool, well done! Is looking fantastic! 🎉 

Found some FancyRadioControl instances that should be ToggleSwitch:
Captura de pantalla 2021-09-29 a las 10 56 17

As a comment to think about:
After having a look on [TransformControl changes](https://github.com/yeahcan/maxi-blocks/blob/0b9bf48613b4441a4cf70ed5c6bda58b25031ad7/src/components/transform-control/index.js) we may need to define the objectives of some components. Now we have:
1. ToggleSwitch:
Replaces the switch responsibility FancyRadioControl had before. It looks like:
Captura de pantalla 2021-09-29 a las 10 01 56

2. FancyRadioControl:
Captura de pantalla 2021-09-29 a las 10 02 38

3. SettingTabsControl:
Captura de pantalla 2021-09-29 a las 10 03 01

4. RadioControl:
Captura de pantalla 2021-09-29 a las 10 03 25

ToggleSwitch has its own responsibility, so no problem. But the difference between SettingTabsControl, RadioControl and FancyRadioControl is so tight. MaxiTabsControl can scape to this filter as it can be annoying in some situations for the way it works as it has full control of the content and the tabs. But the difference between FancyRadioControl and RadioControl itself is tight. IMO FancyRadioControl has lost its essence as has not much difference to RadioControl in terms of code.
So, if there's no other reason to keep this 2 components separately, let's join into one and unique component and avoid duplicate code that becomes double work on maintenance 👍

Naaaaiix (commented):
My comment has no answer, please, share your thoughts about it and how to deal with it 👍 

> As a comment to think about: After having a look on [TransformControl changes](https://github.com/yeahcan/maxi-blocks/blob/0b9bf48613b4441a4cf70ed5c6bda58b25031ad7/src/components/transform-control/index.js) we may need to define the objectives of some components. Now we have:
> 
> 1. ToggleSwitch:
>    Replaces the switch responsibility FancyRadioControl had before. It looks like:
> 
> Captura de pantalla 2021-09-29 a las 10 01 56
> 
> 1. FancyRadioControl:
> 
> Captura de pantalla 2021-09-29 a las 10 02 38
> 
> 1. SettingTabsControl:
> 
> Captura de pantalla 2021-09-29 a las 10 03 01
> 
> 1. RadioControl:
> 
> Captura de pantalla 2021-09-29 a las 10 03 25
> 
> ToggleSwitch has its own responsibility, so no problem. But the difference between SettingTabsControl, RadioControl and FancyRadioControl is so tight. MaxiTabsControl can scape to this filter as it can be annoying in some situations for the way it works as it has full control of the content and the tabs. But the difference between FancyRadioControl and RadioControl itself is tight. IMO FancyRadioControl has lost its essence as has not much difference to RadioControl in terms of code. So, if there's no other reason to keep this 2 components separately, let's join into one and unique component and avoid duplicate code that becomes double work on maintenance 👍

PR #2173, opened 2021-09-23
Issue 1977

elzadj:
Fixes #1977 

Ok, added code that basically checks if we have only one Maxi Default SC, and then applies it on the first load of the Editor.

Advantages of that method: works every time we do wp.data.dispatch('maxiBlocks/style-cards').resetSC(), remove them directly from the DB, install the plugin for the first time, etc.
Disadvantages: the code is kinda stupid. We need to w8 until the store is ready and then check the current SC key and total SC count. And apply SC if Maxi Default is all we have.

Please deactivate and activate the plugin to remove sc-vars option from your DB.

There is another bug with SC reset buttons in the SC Editor -  something in code changes defaultStyleCard values along with the styleCard, which should not happen. Checking wth.


Naaaaiix (approved):
Good one 👍

Alejandro9kgm (approved):

Naaaaiix (approved):

Alejandro9kgm (approved):

PR #2199, opened 2021-10-04
Minor test update for image block

Alejandro9kgm:


javadsamiee (commented):
I think for this part 
`		// size, line-height, letter-spacing
		const inputs = await accordionPanel.$$(
			'.maxi-advanced-number-control .maxi-base-control__field input'
		);
`
please be sure you are in M responsive mode, you selected M, because changing screen size automatically change to another stage, and maybe the result will not be in the M stage
image

Naaaaiix (approved):

javadsamiee (approved):

PR #2218, opened 2021-10-14
fix and update alignment-control

Alejandro9kgm:


javadsamiee (approved):

Naaaaiix (commented):
Well done, good job!

Let's do some details. Move text alignment test to the toolbars test (as this current test is done in the toolbar) on `e2e-tests/components/toolbar/components/alignment` and then create here tests that check the sidebar settings. This way we've got the test on the toolbar and the test on the sidebar, ensuring both are working correctly. Thanks 😊

Naaaaiix (commented):
Have a look as this PR has some failed tests 😄

Naaaaiix (commented):
Last fix brakes the test, please @Alejandro9kgm fix it 👍

Naaaaiix (approved):

PR #2238, opened 2021-10-20
Issue 2234

Naaaaiix:
Fixes #2234


elzadj (changes requested):
Background Shape layer has the same bug - maybe check them all. :+1:

javadsamiee (approved):

elzadj (approved):

PR #2243, opened 2021-10-22
Issue 2225

javadsamiee:


Naaaaiix (changes requested):
The original issue is not completely fixed:

https://user-images.githubusercontent.com/50477222/138552788-5baababc-fc85-4468-b9cf-c6f550757279.mp4

Some integral tests are broken, so this implementation is breaking other things.

Naaaaiix (approved):

Alejandro9kgm (approved):

PR #2247, opened 2021-10-25
Issue 2216

Naaaaiix:
Fixes #2216

### DIFF
--- a/src/blocks/image-maxi/save.js
+++ b/src/blocks/image-maxi/save.js
@@ -16,6 +16,7 @@ import MaxiBlock, {
  * External dependencies
  */
 import classnames from 'classnames';
+import { isEmpty } from 'lodash';
 
 /**
  * Save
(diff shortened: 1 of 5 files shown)


elzadj (changes requested):
Works great now, just one thing: we don't need ```figcaption``` on frontend when it's actually empty (it also adds an unwanted margin): 

Screenshot from 2021-10-25 15-43-45

elzadj (changes requested):
Thanks @Naaaaiix 

I think that if we allow setting up the Caption's setting before we add the image, we should show the Caption placeholder on backend, or we do not really see the changes we make:
Edit-Page-‹-Local-Wordpress-—-WordPress (62)

And a small thing - the alignment here is not right, it's centered by default: 
Edit-Page-‹-Local-Wordpress-—-WordPress (61)

elzadj (approved):

javadsamiee (approved):

PR #2249, opened 2021-10-26
Fix tests 2197

Alejandro9kgm:


Naaaaiix (commented):
Thanks @Alejandro9kgm ; next time, please, ensure to write a description on the PR to let us know what to expect. In this case there are updates to fix #2207 , but also theres a new util for sidebar settings when using e2e tests and there are some skip tests. Without that information is harder to do a review 🙏 

Going to use the tests to compare changes done in #2207 , so this questions are for you @javadsamiee :
1. Container Maxi doesn't contain `alignfull` class anymore. In case is not necessary remove the generated CSS.

2. Column Maxi doesn't contain `data-align`. Not sure if is necessary as it depends on Row Maxi; but, is it an expected result or is an uncontrolled effect?

javadsamiee (approved):

PR #2254, opened 2021-10-26
Issue 2226

Naaaaiix:
Fixes #2226

Points to #2235 👍


elzadj (changes requested):
Great idea! A couple of notes:

- The parallax effect works only on Container Block, not on Text or Row, etc. 

https://user-images.githubusercontent.com/2401618/139023399-b5f3c296-634c-4238-8257-5d6e420d0229.mp4


- The parallax animation looks weird - see the prev. video too - is that expected? :thinking: 


https://user-images.githubusercontent.com/2401618/139024043-0f0bb56f-7949-4a12-a516-c4f24f2d0bab.mp4

- I think we should show some BG Image's settings in the Parallax tab too - like the size and position. People are not going to guess that they need to switch to Normal tab to change it to Cover, and go back to parallax tab again.

https://user-images.githubusercontent.com/2401618/139024519-754f9bdb-f471-46b2-a0b6-64bec385a79d.mp4

elzadj (approved):
Great, thanks!

javadsamiee (approved):

Alejandro9kgm (approved):

PR #2256, opened 2021-10-27
Issue 2222

Naaaaiix:
Fixes #2222

Couple of things:
1. VW is working good, what doesn't work correctly is `em` as resizable-box library doesn't allow it, so removed `em` option.
2. As the maximum size the SVG could get was limited to be responsive and to adapt easily to small screens, added a new option labeled `Force responsive` (change if necessary, is just an idea/concept). It forces the SVG to be adaptive in case is enabled, and leaves free size choosing if creators need/want 👍 

Onces this branch will be merged, will generate an issue to let Alejandro do the SVG tests 👍

### DIFF
--- a/src/blocks/svg-icon-maxi/edit.js
+++ b/src/blocks/svg-icon-maxi/edit.js
@@ -121,14 +121,19 @@ class edit extends MaxiBlockComponent {
 
 		const isEmptyContent = isEmpty(content);
 
-		const handleOnResizeStart = event => {
+		const handleOnResizeStart = (event, direction, elt) => {
 			event.preventDefault();
+
+			elt.querySelector('svg').style.width = 'auto';
+
 			setAttributes({
 				[`svg-width-unit-${deviceType}`]: 'px',
 			});
 		};
 
 		const handleOnResizeStop = (event, direction, elt) => {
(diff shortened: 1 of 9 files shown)


javadsamiee (commented):
Good job 👍🏻

Some points:

1- the maximum value of VM should be limited to 100, now we can have 999
2- When we have for example 563px and change unit to % the value show 563% while the maximum value for % is 100
3- Is it correct when we have % unit for value of SVG width, users resize it with handles all values convert to px?

javadsamiee (approved):

elzadj (changes requested):
Much better, thanks.

The drag n drop resize now looks weird when I make svg bigger - like it resizes horizontally only: 

https://user-images.githubusercontent.com/2401618/139074883-6ec31118-757d-49e4-b3de-64863bb063d8.mp4

I've checked on master - it looks ok, so should be related to the newest changes. If not - make a new issue pls :+1:

https://user-images.githubusercontent.com/2401618/139075400-8cf83cdb-be88-481b-8858-4bacfcbf851d.mp4

Alejandro9kgm (approved):

elzadj (approved):

PR #2272, opened 2021-10-29
Issue 2240

Naaaaiix:
Fixes #2240

- Fix changing image when shape is active and ensure no unnecessary SVG code (when adding a new image, SVG were keeping the old `pattern` element)
- Abstract SVG methods and reorganized in the folder
- Removed unnecessary attributes from Image Maxi and from background component


elzadj (changes requested):
Works great, but it looks like we are missing the same settings for Background Shape Layer: 
Edit-Page-‹-Local-Wordpress-—-WordPress (63)

Compare with the master: 
Edit-Page-‹-Local-Wordpress-—-WordPress (64)

Alejandro9kgm (approved):

elzadj (approved):

PR #2299, opened 2021-11-04
Issue 2277 2268

Naaaaiix:
Fixes #2277 
Fixes #2268 
Fixes #2296 

Make Border number inputs show on responsive the previous responsive value as placeholder, and when clicking/pressing arrow add over the placeholder value; if a number is pressed directly or pasted gets that value (related to #2268 )
Make Typography number inputs show the 'default' value (that comes from SC) as placeholder, so allows removing and interacting with it (related to #2277)
Make Opacity values look all the same (can't be empty, if empty becomes 0), and for it, changed all opacity instances to work under OpacityControl (related to #2277 and as side effect fixes #2296 )
Fix and update tests


elzadj (changes requested):
- The border's dwopdown does not work too - I need to switch from none to any other border first, then it works:

Edit-Page-‹-Local-Wordpress-—-WordPress (66)

- Container has a double border - one for the block, another for the bg displayer: 
Edit-Page-‹-Local-Wordpress-—-WordPress (67)

elzadj (approved):

Alejandro9kgm (approved):

PR #2303, opened 2021-11-04
New scroll motion effects

elzadj:
Not for merging, just for tests. Also pls ignore all the comments and logs in code.

See https://github.com/yeahcan/maxi-blocks/issues/2213 

What is not ready yet:

- added to Image Block and Button Block only for now
- responsive for motions
- frontend motion scripts are not included properly 
- conflicts with other settings (opacity, transform, etc.) are not processed yet
- old motion's code is not removed from everywhere

Any ideas and suggestions of how to make them better are welcomed.  

P.S.: run 'npm install' pls


Naaaaiix (approved):
Great job, congrats 🎉  in terms of working, works reaaaally good!

Just found one error:

https://user-images.githubusercontent.com/50477222/140511424-08086a9c-1911-4d68-beeb-fd37ea4395b9.mp4

In terms of coding: I think this implementation has made a really good base; I would suggest to refactor a bit to make it more solid to ensure an easy access to maintain in future. If we can reduce some methods like `motionData`, integrate the motions on `MaxiBlock` or `MaxiBlockComponent` instead of every block and we fix the conflict that we may find between the commented imperative/declarative actions, motions in Maxi should be a game changer 💪 

Onces finished this branch I would recommend a full list of tests for @Alejandro9kgm 👍

Alejandro9kgm (approved):

Naaaaiix (approved):
Bravo! 👏 👐 🥳 

Probably one of the most pleasant PRs I've ever read, and I would say one of my top ten favorite implementations on Maxi; congrats, your job is spectacular 😸 

As is a faraonic job, is normal to find points to improve or where we disagree; going to share, do the ones you decide or consider, or derive them into new issues and do them later:

- Scroll effect class is pure art, lovely. Super easy to read and to understand, consistent, solid and powerful. It works better great, but I'm afraid about the performance, as I saw my computer get a bit stuck with the animations when I just put one image (don't want to think when creators will put dozens and they will be rendered on an old computer). So, couple of suggestions that can make the class performance enhance:
1. Instead of requesting for the scroll effects data to the HTML element every time we scroll, save the data into the inner class and avoid doing that requests. An HTML element request (`querySelector`, `getElementsBy...`, etc) has a delay that we can avoid just having the static data saved into `this` object variable of the class. That should reduce lots the times on the lifecycle of the class 👍 
2. Apply a viewport checker, something like [`getIsInVierport` used on Parallax class](https://github.com/yeahcan/maxi-blocks/blob/02ee65deac722744805c0214d0f3dc9cfc77eeac/js/front-scripts.js#L160): this can reduce the amount of energy employed when the element is not on the viewport.
3. Related with 1: if the information is saved on the class, it can be removed from the HTML code, which is not a big deal, but will leave a cleaner space if creators want to check.

- About ScrollEffectsControl: have to confess I'm not such a good one to experiment with it as I've never used scroll effects on web builders, so I'm not familiarized with the settings and so on. So all my comment come from my ignorance and probably I'm doing wrong:
1. Not sure if there's any reason to have the preview for every type, but looks more easy for me to have a preview that enable all or no one. I mean, if a creator sets 3 different scroll effect types that will be showed on frontend, enabling/disabling one by one can generate confusion. Maybe I don't understand that logic and is also correct, but came to me as a first impression  👍 
2. Couldn't make the preview work on the backend. Is it working or may I'm doing something wrong?
3. Tried to understand from the settings sidebar and from the code the functionality of `Shortcut effect` setting, but didn't. What's the aim of it?
4. About attributes, I have to say they are hard to read for me, so I would like to suggest the change of the structure. Happened to me that I was trying to read what attributes I've changed on `rotate` and found out that the structure sets the `type` as the third word, so had to go one by one to check what were affecting the `rotate`. Finally decided to copy/paste them into another document and do a search, but that is not comfortable. Maybe there's a reason to set them in the way they are and I didn't noticed, but as I understand the rest of Maxi attributes, the idea is to group them at the beginning of the name to make them easy to read (probably there are attribute groups mistaken, obvios lol). So, if you consider, I suggest to change from an structure like [`scroll-${attr}-${motionType}-${breakpoint}`](https://github.com/yeahcan/maxi-blocks/blob/3b3fc91cae9d188cc8a1f40f70caa9c67a431e5b/src/extensions/styles/defaults/scroll.js#L30) to an structure like `scroll-${motionType}-${attr}-${breakpoint}`.
5. Probably will have new improvements in future, but I've missed to have the shortcut selected appearing as selected on the selector.

- The function `getMaxiBlockAttributes` on `src/components/maxi-block/index.js` is becoming bigger and the file is starting to be really long; we may abstract it on `src/extensions/maxi-block` folder.

Even with that said, going to approve the PR as is working pretty good and what I've wrote are improving suggestions that can be implement in future. Thanks for such a good job @elzadj 😻

Naaaaiix (commented):
@elzadj can you have a look? Tested several times in different ways. Tried adding from shortcut, and does nothing; and then activated manually and nothing happens on frontend. Checked and the class is doing actions on scrolling, but nothing renders. May I'm losing something; thanks 😄

Naaaaiix (approved):
Well done! 🎉 

Main frontend class seems faster, or at least not sticking 👍 Attributes are much much easier to read, thanks so much also.

By what concerns to the code part, it can be merged. I think is necessary that someones reviews each of the effects and if they work as expected before merging 👍 😄

Alejandro9kgm (approved):

PR #2307, opened 2021-11-05
Add anchor option for blocks - Issue 2232

elzadj:
I'm not a fan of changing the actual id of the block to something custom, because we use that uniqueID a lot (style generation, etc.).

Since we need the extra id setting mostly for anchors, here is a test implementation that adds an empty span with the extra id.

### DIFF
--- a/e2e-tests/components/custom-label/index.spec.js
+++ b/e2e-tests/components/custom-label/index.spec.js
@@ -16,7 +16,7 @@ describe('CustomLabel', () => {
 		const accordionPanel = await openSidebarTab(
 			page,
 			'advanced',
-			'add css class id'
+			'add css class'
 		);
 
 		await accordionPanel.$eval(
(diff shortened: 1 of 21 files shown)


Naaaaiix (commented):
aaaaaaaaanndd... if we make `uniqueID` the attribute that can be changed on the ID setting? I mean: the `<span>` is something I'm not sure is going to work. People put IDs to refer with CSS/JS to concrete HTML elements, so the span won't be a way to deal with it. I'm sure more than one will complain about it, as is not the normal and expected behavior. But, we can use `uniqueID` attribute for it. 

`uniqueID` is a must for the style API, but it can be changed and the process should keep the same. Here's a video:


https://user-images.githubusercontent.com/50477222/140543566-709f38d9-3c69-4a97-abf0-d58101294992.mp4

Maybe is not necessary to show the default `uniqueID` on the ID setting, as is kind of ugly, but in case is empty we generate one, in case is filled, we use that one. What do you think?

Naaaaiix (approved):
Cool, it works fine 👍 

Must say I don't understand the reason why we didn't use an ID with `uniqueID` attribute. I'm still thinking is going to be confusing for users, and the implementation would be really easy; but I guess there are some other reasons I ignore to implement this way 🤷 anyway, it works 👍

Alejandro9kgm (approved):

PR #2308, opened 2021-11-08
Update background/layer tests

Alejandro9kgm:
update background / layer and fix loader-control


Naaaaiix (commented):
Found some things to improve and also there are some new tests to add after merged #2310 ; so going to put this PR as a draft and going to list all the tests is necessary to do 👍

Naaaaiix (commented):
Well done! 💪 

Made some improvements, please check them all and ensure you'll be applying on future tests. Added a remover of background layers, so please implement it every time you finish with a layer; it will keep the test cleaner. 

Snapshots are still broken and some comparisons are not working correctly; set some comments on the points it fails. But in general is good 👍

Naaaaiix (commented):
Just one last detail: add `getBlockStyle` to every background layer test to ensure uncontrolled changes on styles.

Naaaaiix (approved):
Good job @Alejandro9kgm 😄

PR #2310, opened 2021-11-08
Issue 2229

Naaaaiix:
Fixes #2229

Implementations:
1. Create a second array for background layers specific for `hovers`. So, all bg layers added to normal bg will appear on hover, but just hover bg layers will appear on hover bg:

https://user-images.githubusercontent.com/50477222/140761428-c8443c73-961b-464e-bc4e-1376943e24d4.mp4

This way we keep separate the layers, so no layer from Normal state can be delete from Hover state. All normal bg can have its version on hover, but just hover layers can be modified from hover settings.
2. Make layers show the correct settings depending the state. For example, a normal video bg layer will show the URL or the `start time` options when creator checks Normal state bg, not in hover. If the layer is added on hover, so just found on Hover state, it show the settings there.
3. Clip-path error and other fixes has been done (#2229 )

@Alejandro9kgm onces merged this branch, ping me to check the tests list 👍


elzadj (approved):
Works much, much better now, thanks.

Alejandro9kgm (approved):

javadsamiee (approved):

PR #2311, opened 2021-11-09
Implement issue 2245 structure (AxisControl)

Naaaaiix:
Implemented the structure for the issue #2245 ; @javadsamiee will take care of the styles and @Alejandro9kgm will start fixing the tests.

Changes done:
1. Changed AxisControl structure. Tried to keep the existent methods and flow as much as posible, but will need a test. The good part is that the new structure and flow is much easier and makes the component simpler.
2. Changed AdvancedNumberControl to allow `auto` option.
3. Implemented BaseControl wrapper to ButtonGroupControl, which allows generating structures with a label on left side and the button groups on the right side.
4. Implemented these changed to Margin and Padding (which also affects iconPadding, etc...), and also to Position and background SVGPosition 👍

### DIFF
--- a/src/components/advanced-number-control/index.js
+++ b/src/components/advanced-number-control/index.js
@@ -44,6 +44,7 @@ const AdvancedNumberControl = props => {
 		value,
 		onChangeValue,
 		disableReset = false,
+		enableAuto = false,
 		onReset,
 		allowedUnits = ['px', 'em', 'vw', '%', '-'],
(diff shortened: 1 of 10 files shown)


javadsamiee (commented):
Good job 👍🏻 

I think a minor thing is a little annoying, when I use the reset button in each part it will return me to the first part, and it should be better to be in the current part I am editing

PR #2315, opened 2021-11-09
Issue 2306

Naaaaiix:
Fixes #2306


elzadj (approved):
Looks good!

Is it possible to not generate the double set of backgrounds? I'm just thinking about custom css for the main block. Need to put any border-related css into two boxes - one for the block, second - for the bg displayer. (not that creators should do it - they have settings for that, but still)

Alejandro9kgm (approved):

javadsamiee (approved):

PR #2318, opened 2021-11-09
Issue 2317

Naaaaiix:
Fixes #2317


javadsamiee (commented):
I think after reset it should be returned to 100 not remain in current value, now the color resets but the opacity value remains.

elzadj (changes requested):
Now the reset button does not do anything.

elzadj (approved):

javadsamiee (approved):

PR #2335, opened 2021-11-17
Issue 2232 - custom css

elzadj:
Adds Custom CSS options to Block - Sidebar  - Advanced. Syntax highlight and css validation included.

Edit-Page-‹-Local-Wordpress-—-WordPress (73)

Notes: 

- **Added to a Button Block and a Map Block only for now.**
- Code for labels generation is meh - we will most likely change them several times, so let's do it manually for now.
- Please do ```npm install``` - you need at least two new packages: ```@uiw/react-textarea-code-editor``` and ```w3c-css-validator```
- There is an idea to add Before and After under one category 'Pseudo' https://www.w3schools.com/css/css_pseudo_elements.asp , but I'm not sure it's good for UI/UX


kyrapieterse (approved):
awesome

Naaaaiix (commented):
Awesome! 👏 Wouldn't done it better. I really love the UI/UX, is easy, clean, powerful, intuitive and works really good. Congrats, as was a difficult one 😸 

I guess is not implemented to all blocks to check before doing a big implementation. I've found couple of basic things I suggest to refactor before continuing that should make future implementations much much more easy:
1. `src/components/inspector-tabs/inspector-custom-css.js` should be lighter, in much senses. First of all, should be really simple: the idea of the inspectorTabs components is to reduce the lines on `inspector.js` files to be easier to read. Also makes easier and faster resolving issues and doing changes that affects all blocks. So, related with it, I suggest to abstract the inner content to create a real component and then the component instanced on the inspectorTabs file; also will make the CSS put on `src/components/base-control/editor.scss:21-75` and `src/css/editor.scss:201-220` to have a correct place. Secondly, the responsibility of the classes or targets or selectors (as is used on the file) shouldn't be on the component, should be on the block, for various reasons: the file will be unreadable large, for example. This is an interesting topic, which continues on next point...
2. Use an object structure to generate the component and to render styles: as is done on the implementation, there are some variables (`selectorsCanvas`, `selectorsButton`, etc...) which are getting the corresponding values depending the block and then are used to create the necessary components. Then, when rendering the styles, `getCustomCss` or `getCustomStyles` functions are looking into the attribute for a concrete 'category' and then for an 'index'. That's kind of confusing in my opinion, and we should be thinking that with the long list of blocks we'll have we'll need to instance this functions every time. So my suggestion is to reduce drastically the amount of instances of that functions and to make the process much more easier and readable. My idea is create an object from the block that corresponds to some kind of standard for CustomCSS component and could look like:
```
// For Button Maxi (abstracted from `src/components/inspector-tabs/inspector-custom-css.js`
{
	selectorsCanvas: ['', ':hover'],
	selectorsButton: [
		'.maxi-button-block__button',
		'.maxi-button-block__button:hover',
	],
	selectorsContent: [
		'.maxi-button-block__content',
		'.maxi-button-block__content:hover',
	],
	selectorsCanvasBackground: ['', ':hover'],
	selectorsIcon: [
		'.maxi-button-block__icon',
		'.maxi-button-block__icon svg',
		'.maxi-button-block__icon svg > *',
		'.maxi-button-block__icon svg path',
		'.maxi-button-block__icon:hover',
		'.maxi-button-block__icon:hover svg',
		'.maxi-button-block__icon:hover svg > *',
		'.maxi-button-block__icon:hover svg path',
	],
	selectorsCanvasBefore: [':before', ':hover:before'],
	selectorsCanvasAfter: [':after', ':hover:after'],
	selectorsButtonBefore: [
		'.maxi-button-block__button:before',
		'.maxi-button-block__button:hover:before',
	],
	selectorsButtonAfter: [
		'.maxi-button-block__button:after',
		'.maxi-button-block__button:hover:after',
	],
}
```
So, with this object send to CustomCSS component from the block's `inspector.js` file, we can generate the necessary components inside CustomCSS; this way the component is not responsible of the CSS targets. With it we can also reduce the amount of times we use the CustomCSS helpers on `style.js`. Let's think we modify CustomCSS to save the attribute object instead of using `category` and `index`, using directly the CSS 'target:
Current `customCss` attr:

```
{
    "button": {
        "0": "background-color: blue; border: 2px solid red;",
        "1": "border: 20px dotted pink;"
    }
}
```

Suggested `customCss` attr:


```
{
    '.maxi-button-block__button': "background-color: blue; border: 2px solid red;",
    '.maxi-button-block__button:hover': "border: 20px dotted pink;"
}
```

So, with this suggested object we don't need to set all the helpers on `style.js`, we just need to take this object, take `response` object from `style.js` and merge each one of the objects with the entries of `response` and in case it doesn't exist on `response`, we add. Finally, with just 1 action that will be the same on all the blocks, we'll be introducing CustomCSS. Also, targets like `':before': getCustomStyles(props, 'before canvas', 0)` which are not inherent to the block styles and will be necessary to add to every `style.js` file will be included in this methodology.

Ping me if any doubt or I can help with something 😄

Alejandro9kgm (approved):

Naaaaiix (approved):
Magnificent, fantastic job! 👌 👏

Couple of comments to improve a bit, but it works and looks fantastic 👍

Naaaaiix (approved):
💯

Alejandro9kgm (approved):

PR #2338, opened 2021-11-18
Small UI fixes

Rustam198804:
Fix issue #2324


kyrapieterse (approved):

elzadj (approved):
```data-name={lowerCase(item.label)}```

So if we rename something, the styles will break. Any ideas @Naaaaiix ?

Naaaaiix (commented):
Well done @Rustam198804 👍

Alejandro9kgm (approved):

Naaaaiix (commented):
Keep advancing, good job 😺 👍 

Most of the issues are related to CSS targets. There are 2 mainly things why we request to change it:
1. Basing CSS targets on `data-name`, which is finally the name of the tab, can be conflictive if in future we change the tabs label. If, for example, `dimensions` tab change its name to `abracadabra`, we'll need to change the CSS. If we can avoid it, is less things to think about and care in future 👍 
2. Every style should go inside its own component, and is important to use the components classes. This helps a lot when need to fix or improve the code.

Thanks for such a good job @Rustam198804 , keep pushing like that 💪

elzadj (approved):
@Rustam198804 
Found this, looks like a small bug: 

Add-New-Page-‹-Local-Wordpress-—-WordPress (3)
Add-New-Page-‹-Local-Wordpress-—-WordPress (2)

Naaaaiix (commented):
Gorgeous 😺 

Just a last small thing to change, and will be ready.  Well done @Rustam198804 , good job and thanks for the patience 🙏 😄

Naaaaiix (approved):

PR #2348, opened 2021-11-22
#2342 Fix height everywhere

Marko64:
Fixes #2342 

Removed the following option:
removed-height


Naaaaiix (commented):
There's a good and a bad new:
1. The good one: you've made your first PR, your first implementation and is everything perfectly good! 😌 Congrats, you're in the game 🤾‍♂️ 
2. The bad one: the solution wasn't the one wrote up on the issue 😅 removing wasn't the real solution, as it was failing for another reason:

https://user-images.githubusercontent.com/50477222/142900062-01cf9941-280f-4b0a-ae0c-26f539df1165.mp4

As you can see the real error is related with the height unit. So, let's keep playing 🏀 👍

elzadj (changes requested):
Aha, understood @Naaaaiix , thanks.

I can't change the unit, and the component breaks after (also: 22.22 px :smile_cat: ):
Edit-Page-‹-Local-Wordpress-—-WordPress (89)

elzadj (approved):
Works, thanks @Naaaaiix 

The setting is not responsive tho - I guess we need it? @kyrapieterse ?

Can be a new issue if we need it.

elzadj (approved):

Alejandro9kgm (approved):

PR #2352, opened 2021-11-23
Issue 2339 - Bottom resizer handle styles on hover

Marko64:
Fixes #2339 

Bottom resizer had a weird animation on hover. I've made it a horizontal line instead, that looks better.

I'm aware this change affects all blocks; however, I can't find any other block using the bottom handle, and Divider seems to be the only block.

### DIFF
--- a/src/blocks/divider-maxi/edit.js
+++ b/src/blocks/divider-maxi/edit.js
@@ -129,10 +129,7 @@ class edit extends MaxiBlockComponent {
 				}}
 				showHandle={isSelected}
 				enable={{
-					bottom: true,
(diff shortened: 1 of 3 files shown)


Naaaaiix (approved):
In terms of code LGTM; not completely sure about the UX when dragging from the bottom, as I don't see how creators will understand they can do that action 👍 

Anyway, @Marko64 you're closer to be part of Maxi code 🤪 well done!

elzadj (approved):
I think it's ok for now, we may add some arrows later. Great work @Marko64 !

PR #2354, opened 2021-11-23
Issue 2049 - Button hover SC settings

Naaaaiix:
This PR fixes #2049 

The implementation has made different and several changes:
1. Removed hover CSS styles from SC settings responsibility from `maxi-themes.scss`. Now it depends on the helpers of the same block.
2. Abstracted 'background' as a new inspectorTab and adapted 'blockBackground' inspectorTab.
3. Enhanced `receiveStyleCardValue` to work with multiple requirements (now supports string/array).
4. Added 'Apply to all [...]` button on hover options of SC.
5. Added different `hover-status` attributes for `background` and `blockBackground` (were using the same).
6. Added support to blocks to connect with its related SC settings and enhanced `maxiBlockComponent` class to re-render when that SC settings updates automatically.
7. Removed `isButton` prop from multiple components and replaced with `globalProps` which puts the responsibility of the SC attributes relation on the block and not the components.
8. Fix tests

There are still some UI/UX situations we need to be aware and to fix. Not a problem at all, just need to decide how to deal with. For example, when we enable some SC setting and we chose custom colour, how should it looks on block sidebar?

https://user-images.githubusercontent.com/50477222/143025380-80c2ce25-a0d0-4d1b-9a10-40f21204291a.mp4

😸


elzadj (approved):
Works great!

Noticed a bug - it's the same on master, so I guess we need new issue - text decoration line's color is wrong, because we apply the decoration to ``` .maxi-button-block__button ```, and the color to ``` .maxi-button-block__button .maxi-button-block__content```

Edit-Page-‹-Local-Wordpress-—-WordPress (73)

https://user-images.githubusercontent.com/2401618/143040680-8fb82677-72d2-4827-97d0-e4cdfde27e71.mp4

elzadj (approved):

Alejandro9kgm (approved):
works excellent!! congratulations😄

PR #2363, opened 2021-11-29
Util color-cotrol

Alejandro9kgm:


Naaaaiix (commented):
The util looks amazing and will reduce hours and lines of work, thanks!

I think not all the ColorControl instances have been updated with this util, and please, use getAttributes util also to complement this optimization  👍

Naaaaiix (commented):
Implement in the rest of cases please 😸

Naaaaiix (approved):

PR #2379, opened 2021-11-30
Issue 2232 Added Custom CSS tab

Marko64:
Fixes #2232 
- Added Custom CSS tab for all existing blocks
- Fixed getCategoriesCss() function to show custom CSS for options that exist. It was inverted.


Naaaaiix (commented):
Very good, in terms of code looks fantastic. Also have to say @elzadj left it really easy 😜 

Apart of the comments left on code, couple of points:
1. May we should decide if we should put `:before` or `::before`; I prefer second option which corresponds to CSS 3 version.
2. I suspect @Marko64 and possibly @Rustam198804 are not working on a VSC environment with Prettier, Beautify and ESLint. We may do a document with the musts we need and the optional VSC extensions 👍

elzadj (changes requested):
Text Block and some other blocks do not have Canvas, so the labels must be 'Before text', 'After text', 'Custom CSS for text ::before' etc., not '... canvas':
Edit-Page-‹-Local-Wordpress-—-WordPress (84)

elzadj (approved):
Found 2 typos in Text, everything else is great, thank you @Marko64

elzadj (approved):

Alejandro9kgm (approved):

PR #2390, opened 2021-12-02
Toolbar fixes

Naaaaiix:
Fixes background and size settings on toolbar. 

Background colour and fullwidth button wasn't working from the toolbar; check on learndd to see the issue 👍


elzadj (changes requested):
The toolbar's full-width does not work for me - nothing happens (see the video). Also noticed that the sidebar's full-width does not work on frontend - can be a new issue if it's a big fix.

https://user-images.githubusercontent.com/2401618/144827009-13fb283b-df23-4db9-9031-d827d28ae1bd.mp4

elzadj (approved):

Alejandro9kgm (approved):

PR #2395, opened 2021-12-03
Issue 2381 - Attributes changes

Marko64:
Joshua at the keyboard:

What has been done:
1. Cleaned non-necessary attributes
2. Removed non-necessary defaults
3. Create `breakpointAttributesCreator`, `prefixAttributesCreator` (from old `getPrefixedObject`) and `hoverAttributesCreator`. Each one of these functions adds to attributes object keys some properties, like breakpoints, prefix or hover. Tests also created for them.
4. Create `paletteAttributesCreator` and `getPaletteAttributes`, both helpers for ColorControl attributes. The first one ensures all the attributes follow the same system and reduce the amount of necessary lines to write; and second makes easier on some situations, like on style helpers, the request of these schema that normally comes all together. Tests also created for them.
5. Updated `stylesCleaner`, which integrates new filters and its content has been re-ordered. The most important one is a breakpoint repeated value checker, which removes the non-necessary/repeated breakpoint attributes that were rendered on the CSS. This way we've removed the amount of CSS generated on every block; here are some examples abstracted from tests:
Captura de pantalla 2021-12-13 a las 20 26 02
7. Update some helpers and tests

The result:
1. Cleaned blocks in terms of attributes, which reduce the amount of data to save and upload/download on Maxi Cloud.
2. Ensure and facilitate the block attributes structure creation with new methods that apply formats to attributes: breakpoints, prefix, hover and palette.
3. Clean the generated CSS on every block, reducing the generated data exponentially (as more attributes are saved, bigger is the difference of generated CSS that was non-necessary).
4. A better world to live... or not? 🤔 lol

Sorry for the amount of files changed 👍


elzadj (approved):
Great work!

Alejandro9kgm (approved):

PR #2397, opened 2021-12-03
Issue 2367 - Default unit and max width fix

Marko64:
Fixes #2367 
- Changed the default unit from 'px' to '%'
- Removed max-width that was set by default for divider. I believe this is a Gutenberg thing, not Maxi's. Added max-width: unset; in style.scss to achieve this.


Naaaaiix (approved):
@Marko64 1 - 0 tests 👊 lol

Good job 👍

elzadj (approved):

PR #2398, opened 2021-12-05
Issue 2369 - Highlighted maxi row block on click

Marko64:
Fixes #2369 
- Added outline for Maxi row, on click, so you can see it has been selected clearly.
We have a similar CSS for columns


elzadj (changes requested):

elzadj (approved):
@Marko64 do you know why do we have ```!important``` for the outline? What will happen if we remove it, and also remove it from editor.scss for the general blue outline? :thinking:

Alejandro9kgm (approved):

Naaaaiix (commented):
Sorry, probably didn't been as specific as I should on the issue description as expected that the video was clear enough, but this implementation is not fixing the error I tried to describe on the issue. 
The bug is related with the toolbar, concretely with the button that enables the dragging option for the columns. When clicking, it doesn't appear highlighted, letting know creators that the action is enabled. Also, the sides of the column doesn't appear highlighted also, so no understanding that the draggers are there.

Ping me if this explaining is not clear enough to explain the issue 👍

Naaaaiix (commented):
1. When clicking on the button for activate the dragging of the columns, I'm expecting to see that icon on orange to know is activate; please, implement it =)
2. When enabling it, the highlighted line added extends the space and I see an overflow bar on the bottom:
Captura de pantalla 2021-12-07 a las 18 15 04

3. Not sure about the Row highlighted line/outline added is necessary, and not sure about the Column draggers styles. May @kyrapieterse should have the last word about it:
Captura de pantalla 2021-12-07 a las 18 19 10

Naaaaiix (commented):
Challenge succesful! 🎉 

Well done, you did it! It works great 👍  Now, left some comments to ensure the CSS code is where it should and some ideas to reduce the line codes and fix some issues 😄

Naaaaiix (commented):
Last step, we are mostly done 👍 😄

Naaaaiix (approved):

PR #2404, opened 2021-12-07
Issue 2336 text block

elzadj:
Fixes #2336 

Added a temporary solution to see an empty Text Block - a grey dotted border.


kyrapieterse (approved):
nice compromise!

PR #2415, opened 2021-12-09
Issue 2402

Naaaaiix:
Fixes #2402 

- Adds `List position` for chose `inside/outside` (the marker gets in/out of the list item).
- Adds `List gap` for chose the left padding of the list.
- **Extra**: adds `List style` to chose between different list styles


elzadj (changes requested):
Thanks @Naaaaiix !

I get 
```react_devtools_backend.js:4045 Warning: React does not recognize the `__unstableMultilineRootTag` prop on a DOM element.```

when I change P to unordered list.

Also I think we need to set up Marker size with a value, otherwise changing it looks weird:

https://user-images.githubusercontent.com/2401618/146552904-5e293541-94fd-4676-9e39-6f72ef904e1f.mp4

elzadj (approved):

Alejandro9kgm (approved):

PR #2419, opened 2021-12-10
Issue 2347 - Added missing fonts

Marko64:
Fixes #2347 

Updated font.js, and added these four fonts.
Was that all, or I need to do something in addition here?


Naaaaiix (approved):
Works perfect 👍

Alejandro9kgm (approved):

elzadj (approved):

PR #2437, opened 2021-12-17
add Caption test and fix Image alt tag test

Alejandro9kgm:


Naaaaiix (commented):
The implementation is correct, but I would like to request another structure. Instead of testing the caption position and gap inside the same test that checks text settings, use it in a complete new test and check the code editor created with the caption on top and with the caption on bottom, and the same on frontend. The caption position is not a CSS property, is an HTML, so is necessary to check the HTML code to ensure it works correct, that's why I'm suggesting this way 👍

Naaaaiix (commented):
Apply the new Typography utils to the caption test please 👍

PR #2440, opened 2021-12-20
Issue 2428 bg shape

elzadj:
Fixes #2428


Alejandro9kgm (commented):
The responsive test is not correct, as we expect the same value of 100 in all expect.

Alejandro9kgm (approved):

Naaaaiix (commented):
On adding an image, the colour tab and the SVG disappears

https://user-images.githubusercontent.com/50477222/146950445-6604c7a5-c9e3-4b91-a9b0-80e3cb83fede.mp4

Naaaaiix (commented):
Cool, I like the toggle idea, much cleaner and easier. Couple of comments:

1. For a better user experience, may we add a 'state' to keep the image url in case creators decide to put an image, then remove, and come back to it again. What do you think?
2. Didn't we fix the SVG image to fit all the height? Found this, not sure if is a bug:
Captura de pantalla 2022-01-10 a las 13 00 23

Naaaaiix (approved):
Cooooool 😸

PR #2445, opened 2021-12-21
Typography test update

Alejandro9kgm:


Naaaaiix (commented):
Something is not working correctly when, apart from the new values, `e2e-tests/components/typography-control/__snapshots__/index.spec.js.snap` snapshot is different from the original. Please, check `e2e-tests/components/typography-control/index.spec.js` does exactly the same it was doing before. 

Also, clean commented lines please

Naaaaiix (approved):
👏

PR #2447, opened 2021-12-22
A very minor change in the test to make it read better

Alejandro9kgm:


Naaaaiix (commented):
I just see some values changed; what should we expect?

Naaaaiix (approved):

PR #2456, opened 2022-01-05
Issue 2450 button group

elzadj:
Fixes #2450 and some small bugs like 

https://user-images.githubusercontent.com/2401618/148248576-4e4e47e7-80d5-4342-935c-f0bb0915ebfa.mp4


Naaaaiix (commented):
Not sure if there's a reason for removing ButtonGroupControl and replace everything with SettingTabsControl; in that case, maybe next comment is not necessary.

-----------------------------------------------------------------------------

The basic idea of every component (as every function) is one component one responsibility. The issue #2450 was generated after ButtonGroupControl was (unexpectedly) incorporated to Maxi with the next schema:
```
<ButtonGroupControl 
    options={[
	{ label: 'X', value: 'x' },
	{ label: 'Y' value: 'y' },
    ]}
/>
{value === 'x' && <XComponents />}
{value === 'Y' && <YComponents />}
```

The issue was that we were reproducing the same responsibility SettingTabsControl has but with other structure in an uncontrolled way that can generate conflicts as we are facing with the indicators or with styles (as we are not setting the new components into a wrapper as SettingTabsControl does). In terms of frontend and for creators it looks the same.

There are some implementations on this PR replacing ButtonGroupControl that leaves us on the same situation but viceversa. For example: AlignmentControl doesn't fit with SettingTabsControl structure (doesn't have content to show) and ButtonGroupControl would fit much better. Also there are other implementations that doesn't fix the problem, like on TransformControl: SettingTabsControl is used as the old ButtonGroupControl and keeps using the commented old schema with the conditionals instead of the content inside each 'item'. This situation makes also SettingTabsControl acquire more than on responsibility 👍 😸

Naaaaiix (commented):
In terms of code seems good. Seen some style errors, not sure if they were expected, all related with the button line:
1. Sometimes the lines is blue, others is red.
2. Sometimes there's no border on the sides.
3. On Transform, for example, doesn't fit all te space:

Captura de pantalla 2022-01-11 a las 17 12 23
Captura de pantalla 2022-01-11 a las 17 12 37
Captura de pantalla 2022-01-11 a las 17 13 56

Alejandro9kgm (approved):

PR #2471, opened 2022-01-14
Fix SVG tests

Alejandro9kgm:
Last time I edited the tests I made a mistake when adding a svg from the modal. As new Svg have been added the order of the modal has changed and therefore the snapshot fails having changed the selected Svg, now it will always take the same svg and should no longer fail.


Naaaaiix (approved):
Thanks @Alejandro9kgm 👍

PR #2472, opened 2022-01-14
Issue 2462 - Palette colour backup

Naaaaiix:
Fixes #2462 and prepares the terrain for #2466 👍 

Basically sets the colour variables under a new schema that ensures that, in case SC are not loaded on frontend, creators/visitors keep having the expected experience of the website:
````
// Change the current:
[whatever]: rgba(var(--maxi-light-color-4),1);
// for:
[whatever]: rgba(var(--maxi-light-color-4,255,74,23),1);
````

In terms of user experience, there's no difference. Just check on CSS that colour with variables is shown as commented on the previous schema 👍


elzadj (approved):
:+1:

Alejandro9kgm (approved):

PR #2482, opened 2022-01-16
Optimization - DB

elzadj:
Fixes https://github.com/yeahcan/maxi-blocks/issues/2444
Fixes https://github.com/yeahcan/maxi-blocks/issues/2237

Ignore the failed tests.

**For testing, not for merging.**  (for now)

You need to deactivate and activate the plugin for it to create the new tables. And you need to re-save any page with maxi blocks for the data to be saved to the DB.

- Moved styles, fonts and custom meta from _options table to maxi db tables. _What left to move: scroll motions, sc things._
- Chopped front-js into several .js files 
- Loading them conditionally https://github.com/yeahcan/maxi-blocks/issues/2237
- Added uninstall.php to clean up after creator removes our plugin.
- Removed traces of the old motions.
- Fixed some typos and naming inconsistency. 
- Change the Custom Data structure


Naaaaiix (commented):
Oh, much pleasure to start seeing our own DB 😸

Naaaaiix (approved):
Congrats, this job is superb! 👏 🥳 

> I've added a command npm minify:frontend to minify the frontend scripts in /js folder. Not sure if the script will work on Windows tho. And I left the not-minified scripts in that folder too, in case we need to change them. Need to think of something for npm build and the release version.

Yes, we should add it as an action of Webpack; not urgent but necessary for stable version 👍

Alejandro9kgm (approved):

PR #2499, opened 2022-01-24
Redistribution of background layer tests

Alejandro9kgm:
With this PR I have separated the background-layers tests to be able to be checked more easily if an error appears in any of the tests.

Background-displayer and background-control tests have been removed


Naaaaiix (commented):
Please, check the background element inside of the block `.maxi-background-displayer` exists on editor and frontend. Also, in video and shape tests do a snapshot of it in both editor and frontend 👍

Naaaaiix (changes requested):
My last comment maybe wasn't clear enough, so I'll try to divide into different points:

> Please, check the background element inside of the block .maxi-background-displayer exists on editor and frontend. Also, in video and shape tests do a snapshot of it in both editor and frontend 👍

1. Check if the element `.maxi-background-displayer` exists on editor and frontend: no need of snapshot, just need to check it exists. It means:
```
// On the editor
const bgDisplayerEditor = page.$eval('.maxi-background-displayer', el => !!el);

expect(bgDisplayerEditor).toBeTruthy();

// same on preview page
[...]
``` 
2. For video and shape layers, instead of the point one, do a snapshot. In editor and in frontend, both places please.

PR #2513, opened 2022-01-31
issue-2505 is fixed!

FitseTLT:


Alejandro9kgm (commented):
The alt tag option of none does not work correctly.

https://user-images.githubusercontent.com/58038966/151928725-b6fff4f3-32d5-46b2-841f-4bbef762c4d0.mp4

elzadj (changes requested):
The default is still Wordpress Alt, that returns [object Object] when it's empty, or false:

```
react-dom.js?ver=16.13.1:82 Warning: Received `false` for a non-boolean attribute `alt`.

If you want to write it to the DOM, pass a string instead: alt="false" or alt={value.toString()}.

If you used to conditionally omit it with alt={condition && value}, pass alt={condition ? value : undefined} instead.
    in img (created by media_upload_MediaUpload)
    in div (created by HoverPreview)
    in HoverPreview (created by media_upload_MediaUpload)
    in div (created by Resizable)
    in Resizable (created by ForwardRef)
    in ForwardRef (created by media_upload_MediaUpload)
    in media_upload_MediaUpload (created by WithFilters(MediaUpload))
    in WithFilters(MediaUpload) (created by edit)
    in figure (created by ForwardRef)
    in ForwardRef (created by ForwardRef)
    in ForwardRef (created by edit)
    in edit
....
```

https://user-images.githubusercontent.com/2401618/151959131-b1b71a2f-c56e-4329-80f4-30b41f5f59b3.mp4

Naaaaiix (changes requested):
As @elzadj commented, the alt item is not working properly. Here you have a video with some clues about how to fix it:


https://user-images.githubusercontent.com/50477222/151975909-fa90a5b7-4abf-4c29-9d4b-dbc23f794d4e.mp4

elzadj (approved):

Alejandro9kgm (approved):

Naaaaiix (approved):
Cool, the issue has been fixed, good job 😸 

> I have implemented Josh's comments. But I didn't like the part that Wordpress Alt option should be removed from the options whenever there is no wordpress alt for the image, because if a user wants to input wordpress alt after the option is already removed he can't choose Wp Alt option unless the page is reload because it is then that the image alt control component is loaded and the wp alt option is reinserted. Instead, I would prefer to only change to image title option when the wp alt is empty without removing the wp alt option. I think that is more logical!!!

Well, the reality is harder than it looks. The problem is more related with WP than with Maxi, and what you suggest to keep the WP alt option on the select would be a bigger problem than a solution. The main point, as you'll see in the video, is that the store where all this information is not update when selecting the image, so we can't access (using WP native methods) to the update done on the image. Would mean for us to create our own store and do a bigger job of something we may not need for the moment. Leaving the option as you comment, and without the capacity of updating the WP alt would be much more confusing than anything. Thanks for your implication, keep sharing your thoughts as is what makes us bigger and more powerful 😄 

https://user-images.githubusercontent.com/50477222/152368341-f5b7f7b1-d4d0-44d5-a94e-2321c5a71260.mp4

PR #2561, opened 2022-02-11
Issue 2432 - Block buttons are affected by Block settings

FitseTLT:


Naaaaiix (commented):
In general it works good, but is going to be a tricky issue due to the many targets it has that increments the complexity of the task. Here's a video about what I've found 👍 

https://user-images.githubusercontent.com/50477222/154046126-399e4ba1-c509-483a-8c1e-b56885a00bed.mp4

Alejandro9kgm (approved):

elzadj (changes requested):
The svg modal does not work for me, and it's not showing any errors. Works on master tho. Weird.


https://user-images.githubusercontent.com/2401618/161499233-a3f78801-c856-46b6-a964-6dc816d3354d.mp4



Also we need to do something with the Number Counter (move the re-play button to the toolbar?):
Edit Page ‹ Local Wordpress — WordPress (25)

elzadj (changes requested):
I need to click out of the block first to be able to open the modal

https://user-images.githubusercontent.com/2401618/161525605-50c4f89a-32c5-43b9-b510-72f7b9b07481.mp4

Also this looks weird:
Edit Page ‹ Local Wordpress — WordPress (26)


@kyrapieterse is that ok if we move the re-play button to the toolbar, similar to the upload button for an image?

elzadj (approved):
Everything is great, thank you! Just remove the commented css code and MEEEERGEEE

PR #2571, opened 2022-02-15
add list test

Alejandro9kgm:
https://github.com/yeahcan/maxi-blocks/issues/2530


Naaaaiix (commented):
Can you please send the structure we talked on the videocall, please? Because this one doesn't fit with the one I have, and there are thinks not tested. Going to repeat, this time writing it here:
1. Organized list (general things, not start point or reverse)
2. Start point and reverse on organized. Test the minus, test the reverse, test changing to any other type (for example `upper roman`) and check the minus is not available.
3. Unorganized list (general things, not custom types)
4. Custom types of unorganized list

In this implementation I just see 2 tests 👍

Naaaaiix (changes requested):

Naaaaiix (changes requested):

Naaaaiix (approved):

PR #2580, opened 2022-02-16
Text font reset button #2574

FitseTLT:


bilouStrike (approved):

Alejandro9kgm (approved):
works fine, do you think it would be better if it returned your old selection instead of the default selection?🤔

Marko64 (approved):

Naaaaiix (changes requested):
Instead of implement the reset button on TypographyControl, please, add it on FontFamilySelector. That way will be reusable on different situations. Right now, there are places like this one where the implementation scope is not arriving:
Captura de pantalla 2022-03-08 a las 10 35 04

Naaaaiix (approved):
Cool, good job! 👍 

Fixed the tests; onces merged, please @Alejandro9kgm , do a test for it 👍

PR #2581, opened 2022-02-16
add test to reset button in color-control

Alejandro9kgm:
https://github.com/yeahcan/maxi-blocks/issues/2533


Naaaaiix (commented):
In the issue we requested 3 test, I just see one. Is there any reason why they are not implemented? 🤔

PR #2582, opened 2022-02-16
Button Maxi style bug #2554

FitseTLT:
…ad of array in button maxi's style.js


Marko64 (approved):

bilouStrike (approved):

elzadj (changes requested):
Thanks! 
See the tests - border radius for shortcut 2 doesn't work, indeed. Can you figure out why, please?

Naaaaiix (changes requested):
As @elzadj  commented, if we are having a broken test with this implementation there are 2 options: or test was not working correctly (we need to believe they are correct) or something in the implementation run bad. Please, check which of both situations is the actual 👍

Naaaaiix (changes requested):

Naaaaiix (changes requested):
Had a look at the broken tests, and they are right, so can be updated. What we see on them is that there's a part related with the `border-radius` which is removed. This happens because before the implementation `stylesCleaner` function couldn't affect the obj send by `style.js`; now this mentioned method can affect the object, it cleans it out, and removes part of the CSS that is not necessary 👍 

Other thing found, not sure if is related with this issue (if not, derive to a new one):

Captura de pantalla 2022-02-21 a las 16 41 17

Naaaaiix (changes requested):

Naaaaiix (commented):
Please @FitseTLT , update tests 👍

Naaaaiix (changes requested):
Found this CSS code:
`.maxi-button-block__icon svg > *{ fill: none; stroke: rgba(var(--maxi-light-color-1,255,255,255),1); margin-left: 5px;}`

the target `.maxi-button-block__icon svg > *` is affected by margin. It shouldn't be there 👍

Naaaaiix (changes requested):
Captura de pantalla 2022-03-01 a las 13 37 26

🤔

Naaaaiix (changes requested):
Still having this issue:

https://user-images.githubusercontent.com/50477222/156736677-c0707524-9fe7-4a70-8c3e-2386875bcd62.mp4

Alejandro9kgm (approved):

Naaaaiix (changes requested):
Still finding issues with the margin:
Captura de pantalla 2022-03-09 a las 9 28 29
Captura de pantalla 2022-03-09 a las 9 27 52

Naaaaiix (approved):
Cool, works good now 😸 

Fixed tests, onces checked this branch can be merged 👍

PR #2583, opened 2022-02-16
add sc-variable test

Alejandro9kgm:
https://github.com/yeahcan/maxi-blocks/issues/2528


Naaaaiix (approved):
Is normal to need a block, or at least something wrote into a native WP Paragraph block; if not you can't save, and without saving you can't preview lol

Good job 👍

PR #2586, opened 2022-02-17
Issue-#2459 Add ResponsiveTabs breakpoitns

bilouStrike:
Fixes : #2459 

**How to test:**
1- Add your block
2- open style collection(border, alignemnt ....) check if reponsiveTabControl is present in the top of this inspctor accordion.
3- switch to a breakpoint and change the style : check if the indicator is added to the reponsiveTabControl 👍


Naaaaiix (changes requested):
Found couple of error on the implementation, please fix them to keep with the review. 

Checked what you've commented by chat about the error found on indicators. Couple of things:
1. When finding a problem and pushing a PR requesting help for it, please write the problems you're facing on the PR. This way we are all concern about the situation and other devs can also help or give more context about the situation. Also makes much easier understanding the situation and keeping a trace of the conversation and the ideas shared or steps done 😸 
2. In this case indicators error is something not related with this implementation and needs a special treat with its issue and its decision. What mostly happens are 2 situations: first one is that not always the reset button returns the default attribute, which means that even when it looks as default on frontend, the attribute is not the same. Second situation is related with background layers, that will need a special implementation in terms of indicators, as they don't have 'attributes' as we normally use them on blocks, it uses an array of objects with attributes inside, so the mechanism is completely different. 

So, resuming, fix the commented lines of the implementation and leave the indicators issue for fixing after this implementation 👍

PR #2623, opened 2022-02-28
issue-2599 Add # & / to allowed character

bilouStrike:
Fixes: #2599


Marko64 (commented):
Works great, thx. 
But we are missing "." and ":" too. My bad, I should have noticed it right away.

Then we have all the characters needed for a full URL like this one:
https**:**//www.google.com/

Naaaaiix (commented):
Not sure about the issue and the implementation. I think this feature was thought  in other sense, and is to create the reference where we want to anchor. I mean, this field generates the name where to point:


https://user-images.githubusercontent.com/50477222/156167771-d109cbcc-0b0a-4574-a201-28156bedd92d.mp4


🤔

PR #2625, opened 2022-02-28
Issue 2615 - Fix z-index

Naaaaiix:
Related to #2615 && #2616 

The implementation had a bug that generates situations where adding a background hides the content. To reproduce the error from master, create a Text Maxi block, write something and then add a colour background layer: the text is not visible anymore.

To test this fix:
1. Adding a block shouldn't hide the content
2. Adding an arrow into a Group Maxi with other blocks should make visible not just the content also the arrow
3. Adding a second layer, like image one, to the background of the Group Maxi should make the arrow keep the colour of the first colour layer and the block should have the image

An image of the three examples working together:
Captura de pantalla 2022-02-28 a las 13 15 40


Marko64 (approved):
Looks good

Alejandro9kgm (approved):

bilouStrike (approved):

PR #2629, opened 2022-02-28
Issue 2604 - Minor fix

Naaaaiix:
Here's a fix related to [this comment](https://github.com/yeahcan/maxi-blocks/pull/2609#issuecomment-1054156749)

For testing, recreate the same commented before and check it works correctly 👍


Alejandro9kgm (approved):

Marko64 (commented):
Omitting top and bottom margin help with those, but you still can't use side margins for a container. If you go to the container settings > margin - try adding left/right margin. It will not work because of the important rule again.

bilouStrike (approved):

Marko64 (approved):

PR #2646, opened 2022-03-02
Update responsive test arrow control

Alejandro9kgm:


Naaaaiix (changes requested):
Just small thing, the rest is good 👍

PR #2655, opened 2022-03-03
Update divider tests

Alejandro9kgm:


Naaaaiix (changes requested):
The issue was:
```
1.Do the test from sidebar, not from toolbar

2.Test defaults button in one test; test horizontal divider and its responsive in another test; test vertical divider and its responsive in other tests and test the mix of vertical and horizontal divider changing on the different responsive stages
```

So:

- [x] Do the test from sidebar, not from toolbar
- [x] Do one test with just defaults
- [x] Do one test of horizontal divider and its responsive (*)
- [x] Do one test of vertical divider and its responsive (*)
- [x] Do one test mixing vertical and horizontal and changing them on different responsive stages (*)

Just 2 points are done as requested. Please, follow the instructions requested, as there are reasons to ask like that. Also, on the (*) add style snapshot 👍

Naaaaiix (changes requested):
I don't really see the structure asked on the last comment, can you please try to explain how did you test all what has been required?

Naaaaiix (approved):

PR #2696, opened 2022-03-14
issue-2626: Maxi style when removing from a wrapper block

bilouStrike:
Fixes: #2626 

**How to Test:** 
1- Add group-maxi block and define blockStyle: light/dark
2- Add block outside goup-maxi and define blockStyle: different from maxi-group
3- Move the block inside wrapper (maxi-group), it should take maxi-group blockStyle, and when Move it again outside goupe-maxi, it should take, the block original blockStyle.
Please add all possibles cases of changing block and the block wrapper blockStyle property.


Alejandro9kgm (approved):

Marko64 (approved):

Marko64 (approved):
Works as expected now, good job.

FitseTLT (approved):

elzadj (approved):

Naaaaiix (commented):
Good job @bilouStrike 😸 

I'm wondering: with this implementation, is still necessary to keep having `blockStyle` as seems all responsibility is laying under `parentBlockStyle`? I mean, what's the difference now?
In case there's no difference or no reason to keep `blockStyle`, I'd suggest to modify `parentBlockStyle` instances and give `blockStyle` label as seems more accurate 👍

PR #2701, opened 2022-03-15
add test for button icons

Alejandro9kgm:


Naaaaiix (commented):
There are some settings didn't test, like 'Icon position', Icon colour or Icon border. Also please test styles and hover version 👍

Naaaaiix (changes requested):
Still needs the hover version please 👍

Naaaaiix (approved):

PR #2705, opened 2022-03-16
Issue 2420 - Added limits on the range in AdvancedNumberControl

Fybex:
Fixes #2420 

1. Added new prop: minMaxRangeSettings. It similar to minMaxSettings, but contain min, max on range to every unit with default values which was recommended in the thread of the issue.
2. Added disableInputsLimits parameter, which, if true, disable the range.
3. Added new const variables: minValueRange, maxValueRange and changed into them in min/max RangeControl settings.
4. Changed check in units Select control.

How to test:
1. Check slider min/max values in some parameters like width in different components.
2. Switch between units. For example: from 1999px to em. It should be by default set to 199em.
3. Change default minMaxRangeSettings in some blocks.

### DIFF
--- a/src/components/advanced-number-control/index.js
+++ b/src/components/advanced-number-control/index.js
@@ -55,32 +55,54 @@ const AdvancedNumberControl = props => {
 			px: {
 				min: 0,
 				max: 3999,
-				minRange: 0,
+				minRange: -1999,
 				maxRange: 1999,
 			},
 			em: {
 				min: 0,
 				max: 999,
-				minRange: 0,
+				minRange: -199,
 				maxRange: 199,
 			},
 			vw: {
 				min: 0,
 				max: 999,
-				minRange: 0,
+				minRange: -199,
 				maxRange: 199,
 			},
 			'%': {
 				min: 0,
 				max: 100,
-				minRange: 0,
+				minRange: -100,
 				maxRange: 100,
 			},
 			'-': {
 				min: 0,
 				max: 16,
-				minRange: 0,
+				minRange: -16,
 				maxRange: 16,
+			},
+		},
+		minMaxRangeSettings = {
+			px: {
+				min: -1999,
+				max: 1999,
+			},
+			em: {
+				min: -199,
+				max: 199,
+			},
+			vw: {
+				min: -199,
+				max: 199,
+			},
+			'%': {
+				min: -100,
+				max: 100,
+			},
+			'-': {
+				min: 0,
+				max: 16,
 			},
 		},
 	} = props;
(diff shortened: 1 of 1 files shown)


Naaaaiix (approved):
Impressive job, congrats 🥳 👏 

Let me know what do you think about this:
Instead of having 2 different objects (`minMaxSettings` and `minMaxRangeSettings`) for limiting the unit settings, what if we reduce all to same object like:
```
{
	px: {
		min: 0,
		max: 3999,
		minRange: -1999,
		maxRange: 1999,
	},
	em: {
		min: 0,
		max: 999,
		minRange: -199,
		maxRange: 199,
	},
	vw: {
		min: 0,
		max: 999,
		minRange: -199,
		maxRange: 199,
	},
	'%': {
		min: 0,
		max: 100,
		minRange: -100,
		maxRange: 100,
	},
	'-': {
		min: 0,
		max: 16,
		minRange: 0,
		maxRange: 16,
	},
},
```

To make it even more practically when instancing the component we can just modify one of the allowed settings, something like:
```
<AdvancedNumberControl
    minMaxSettings={{px: minRange: 99}}
   [...]
</AdvancedNumberControl>
``` 

so, using a lodash `merge` get the whole object and just modify the requested ones over the default:
```
const minMaxSettings = {
	px: {
		min: 0,
		max: 3999,
		minRange: -1999,
		maxRange: 1999,
	},
	em: {
		min: 0,
		max: 999,
		minRange: -199,
		maxRange: 199,
	},
	vw: {
		min: 0,
		max: 999,
		minRange: -199,
		maxRange: 199,
	},
	'%': {
		min: 0,
		max: 100,
		minRange: -100,
		maxRange: 100,
	},
	'-': {
		min: 0,
		max: 16,
		minRange: 0,
		maxRange: 16,
	},
};

const instance = {
  px: {
    minRange: 99,
  }
};

const result = merge(minMaxSettings, instance);
```

The result would be:
```
{
   "px": {
      "min": 0,
      "max": 3999,
      "minRange": 99,
      "maxRange": 1999
   },
   "em": {
      "min": 0,
      "max": 999,
      "minRange": -199,
      "maxRange": 199
   },
   "vw": {
      "min": 0,
      "max": 999,
      "minRange": -199,
      "maxRange": 199
   },
   "%": {
      "min": 0,
      "max": 100,
      "minRange": -100,
      "maxRange": 100
   },
   "-": {
      "min": 0,
      "max": 16,
      "minRange": 0,
      "maxRange": 16
   }
}
```

Reducing lots the necessary lines and the efforts for implementation 👍 I think it can be so helpful and make the code much much simpler and cleaner. 

Anyway, you made an awesome job as the implementation  is 💯 😸

Naaaaiix (approved):
Awesome 👏 🥳

elzadj (approved):

Alejandro9kgm (approved):

PR #2715, opened 2022-03-16
Issue-2649: ClipPathControl fixes and improvements

bilouStrike:
Fixes: #2649


Marko64 (approved):

Alejandro9kgm (approved):

FitseTLT (approved):

Naaaaiix (changes requested):
Good implementation, it works good, but we might need to simplify it. If every time we need to instance ClipPathControl we have to write this amount of code is going to be insane and a weak point where we'll make mistakes; I mean, half of the props of the component shouldn't be there. As the rest of the components, ClipPath should be able to receive all the `clipPath` attributes using:
```
{...getGroupAttributes(
	attributes,
	'clipPath',
	false,
	prefix
)}
```

and then, combining the breakpoint and the prefix, return a simple `onChange` with all the new attributes. Delegating the responsibility of this conditional to every instance of the component is not a good practice:
```
					onToggleClipPath={val => {
						onChange({
							[getAttributeKey(
								'background-image-clip-path-active',
								isHover,
								prefix,
								breakpoint
							)]: val,
							...(!val && {
								[getAttributeKey(
									'background-image-clip-path',
									isHover,
									prefix,
									breakpoint
								)]: null,
							}),
						});
					}}
```

It should be done on the same component 👍

Naaaaiix (changes requested):

Naaaaiix (changes requested):

Naaaaiix (changes requested):

Naaaaiix (approved):
Fantastic! 🤟

PR #2716, opened 2022-03-16
Add expect to typography reset button

Alejandro9kgm:
Small test to check the Typography reset button is working properly


Naaaaiix (approved):
Please, set a description about what is this PR doing. I had to go to the issue which doesn't give more information than a PR, that I had to read to understand what this test is doing. Would help the team work and the workflow a better description of what is this PR aiming to implement 👍

PR #2723, opened 2022-03-17
Issue-2695 : Add Flex settings

bilouStrike:
Fixes: #2695


Marko64 (approved):
Great! I love this. Flex is great. 
So far everything I tested works well. Played with containers, rows and blocks.

We will have to add a flex guide in our documentation though. We def want to avoid people asking questions how to use it since there are a lot of options.

FitseTLT (approved):

Naaaaiix (commented):
So beautiful, good job @bilouStrike 😻 

Just one minor thing: on cases where the block is not inside any wrapper the Flex accordion is appearing, but empty. It shouldn't appear. Also, I'd suggest in cases like a Column or Group, where we are mixing Flex-parent and Flex-child options, we make kind of difference between both groups. Not sure how, maybe a tab? If you don't have any idea of how to implement it we can merge this implementation and then find a way to do show it 😄 

PS: something really big is happening as this implementation broke practically all e2e tests lol have a look on it please 👍

Naaaaiix (commented):
I miss `order` setting, can you also implement please? 😸

Naaaaiix (changes requested):

Naaaaiix (changes requested):

Naaaaiix (changes requested):

Naaaaiix (changes requested):

Naaaaiix (changes requested):
Something is not working correctly with the helper that breaks the tests:

Captura de pantalla 2022-03-28 a las 11 43 58

So, please:
1. Create a helper test (some examples on `src/extensions/styles/helpers/test`)
2. Fix the helper 😄

Naaaaiix (approved):
Just this small thing and will be merged! 🤟 good job @bilouStrike 😸

PR #2746, opened 2022-03-22
Issue 2699/2475/2534 - New block inserter

Naaaaiix:
Related to #2699, fixes #2475 and fixes #2534

This implementation is a part of #2699 related to the first point:
> New appender component: instead of using the original appender (the plus button) on the bottom of the wrapper blocks, leave the same + button but instead of opening the option to add the blocks on click, show a dropdown with the different nested wrapper blocks and let creators choose. That way we fix #2475 or basically the overlapping of the plus button that makes hard to decide where to add the new blocks. This fix should also fix #2534

The implementation has been done into a popover, which generates the new button as a fixed positioned item outside of the block. There are different reasons to do it:
1. There was a conflict on `z-index` impossible to fix as the inner blocks that contained the appenders would always get on the backwards, so the appender showed was always the first parent one.
2. There was a conflict with the flex: on creating a Column Maxi, for example, having the appender as part of the block and using `vertical-align` setting with `space-between`, the appender was taken as an element, generating an extra space for it that doesn't correspond with the frontend view:

| Backend | Frontend |
|---|---|
| Captura de pantalla 2022-03-22 a las 12 01 23 | Captura de pantalla 2022-03-22 a las 12 03 27 |

Important: this implementation is just a prototype to start working on. Is not styled, is not fancy and may have some error or bad UX. Please, share your thoughts and ideas about it 😸 

To test: this is the main idea, you can play in different ways with it 😄 

https://user-images.githubusercontent.com/50477222/159469576-4ddf08d6-3f0c-411b-8b0e-1b8548516d15.mp4


Alejandro9kgm (approved):

FitseTLT (approved):

elzadj (approved):
Great idea and implementation, but really needs styling

 @kyrapieterse ?

PR #2748, opened 2022-03-22
issue-2737 fixed icon fill issue on button-maxi

FitseTLT:
fixes #2737 
@Naaaaiix Don't you think we should give the users to change the icon fill color on button-maxi? Now, I made it to have the icon's default fill when the icon inherit is false, and when it's true fill will be none. I think it's quite logical.


Alejandro9kgm (approved):

Marko64 (approved):
Works good. Just a different icon here :) 
Screenshot_1

Marko64 (approved):

Naaaaiix (changes requested):
Need to fix the tests @FitseTLT , there are some that just need update, but there are other that seem is showing that we might have an error on the implementation 👍

Naaaaiix (approved):

elzadj (approved):

PR #2753, opened 2022-03-23
Optimization font uploader

elzadj:
Fixes #2526
Fixes #2749 

Please ignore the ugliness of the options.php - Maxi dashboard is an entire different story (and issue) :D
Maxi Blocks ‹ Local Wordpress — WordPress (1)


Alejandro9kgm (approved):

Marko64 (approved):

bilouStrike (approved):

Naaaaiix (commented):
Good job, is a hard task 👍 

Added a Text Maxi. and found that in a default installation Roboto font is required from `wp-content/uploads/maxi/fonts/roboto/style.css` but the file doesn't exist, not sure from where is taking the font. Changed the font to Montserrat, and the same: no file on the folder, and on frontend Text Maxi is not showing the font. Maybe is related to some problem with permissions on the folder on my local as the rest of reviewers approved it (should work, no?), but can be a good point of view for creators that may find same issue than me. How can we deal with it?

Naaaaiix (approved):
Good and big job, congrats @elzadj 😸 

Merge it when 🔥 ✌️

PR #2778, opened 2022-03-29
Issue: 2757 Fixes on Copy/Paste is not working for some props 

FitseTLT:
@Naaaaiix this the mapping idea you suggested implemented on button-maxi only so check it out before it is implemented to other blocks. Thanks!!!


Naaaaiix (commented):
We need to refine the idea:
1. What we seek is an object with all block possible group attributes as an entry, and in each one of this entries we need the key to be a human-readable label to set on the CopyPaste component and the value the attribute group name.
2. We might have an object with general group attributes entries used in all (or mostly all) the blocks as a base, so every block should complete with its specific attributes groups.
3. Each block object should be on a block file, I mean, every block folder should have the file there and should be send, by `edit.js` to the toolbar, where the CopyPaste component should be able to check the group attributes to copy and the ones that are copied and can be paste.

As said on the issue, the component was done fast and to be functional. Now we need some love for it, so don't expect using the base we've got for it. Time to play and to create @FitseTLT , show us you magic 🪄 😸

Naaaaiix (commented):
Well done 👍 

I see that having a general object with the shared group attributes is not as easy as thought, so no need to add it.

Naaaaiix (approved):
Good implementation, this one wasn't easy 😜 fantastic job @FitseTLT 🥳 

For me this branch can be merged on master, but I' like to derive 2 new tasks for it:
1. Abstract `getOrganizedAttributes` and `cleanStyleAttributes` and create unit tests for them as they are really complex functions and we need to ensure they always work. A way to shield this feature 👍 
2. Need to find a way to improve the UI on the groups of attributes, is confusing IMO:
Captura de pantalla 2022-04-21 a las 12 10 35

Alejandro9kgm (approved):

Marko64 (approved):

PR #2784, opened 2022-03-30
Issue 2774 - Drag and drop block moving experience

Naaaaiix:
Partially fixes #2774

- Reduce unnecessary Row Maxi renders
- Reduce unnecessary MaxiBlock component renders
- Fixes new block inserter presence: the inserter was appearing all time.
- Fixes svg icon: it had an error on console

To test: using master, use this [editor code](https://github.com/yeahcan/maxi-blocks/files/8381438/editor.txt) to generate a page with lots of blocks and then try to move some block from one column to other. See how long it freezes. Then do the same with this branch and see if there's a difference (theoretically is there lol).

Still will have some delay, as there are some improvements that can be done, but this is a first part of them and should reduce the freeze effect 👍


Alejandro9kgm (approved):

Marko64 (approved):
This is a tricky one. I do see some improvement, but still feels a bit slow. So, my approval goes to changes on better, but happy to hear that you know there are other things that can be done. Speaking of it, just checked Divi a little. When you start dragging a block, it disappears, and you only see an icon - then the block becomes visible once you drop it. This might be to make the process faster? Perhaps that's an idea.

elzadj (approved):

PR #2787, opened 2022-03-31
Issue 2263 css optimization

elzadj:
Fixes #2263 

You need 

```npm i --save-dev autoprefixer minify-css-string postcss optimize-css-assets-webpack-plugin```

Please make sure that ```npm build``` works, I've checked it on linux only.

Also please check ```npm run minify:frontend``` too - it was updated to work on any system (if you have minify installed globally).

### CODE REVIEW

**Review by Naaaaiix — COMMENTED** [review]
How long dreaming on seeing our CSS minified and prefixed. Amazing job, congrats! 👏 

PS: is still necessary to keep the global dependencies installed? If yes, we might update wiki with it for future devs 👍

**Review by Naaaaiix — APPROVED** [review]

**Review by Naaaaiix — COMMENTED** [review]
Not sure if related @elzadj , but have this on debug.log:
```
PHP Warning:  array_merge_recursive(): Expected parameter 2 to be an array, null given in /var/www/html/wp-content/plugins/maxi-blocks/core/class-maxi-local-fonts.php on line 68
```

and also, may not related, but after remove the fonts using the option on Maxi block options page, I couldn't download any new one 🤔

Naaaaiix (approved):
Seems it works good now, thanks for the job 😸 

Pity the time delay while downloading the font that can show a frontend without the selected font. We might set a loader advice for creators 👍

FitseTLT (approved):

Fybex (approved):

PR #2847, opened 2022-04-11
Issue 2612 - bug in text maxi lists

myroslavve:
Fixes #2612


Marko64 (approved):

FitseTLT (approved):

Naaaaiix (commented):
The implementation fixes the issue, but I have my concerns with the issue itself. Why is a problem to keep the `listStart` negative value saved on the attributes if we are going to set 0? I see next situation: creator sets a negative `listStart`, changes type of list (`listStart` saves as 0) and then creator want to come to first type of list => needs to set the `listStart` again. So, not sure about if it was an issue or not 👍 not also a big deal lol

Other thing; on tests I found this:
Captura de pantalla 2022-04-18 a las 10 47 11

Something is not working correctly on the test when the type of list is `armenian` and the `listStart` is not 0 🤔

elzadj (approved):

PR #2850, opened 2022-04-12
Removed image overflow

Marko64:
.maxi-image-block-wrapper had overflow: hidden;
Must be for a reason, but can't see which one. It was added a long time ago. On the other hand, it was creating issues with border and box-shadow:

Screenshot_1
Screenshot_4

Removed the overflow and tested image sizing, shapes and clip-path. Everything seems good. If there are no objections, going to leave it like this.


bilouStrike (approved):

Naaaaiix (commented):
There's a big improvement on this implementation but still has a weak point:

https://user-images.githubusercontent.com/50477222/163336026-4a3836f1-1dfb-4bf8-b1c3-d95c3e5f8953.mp4

Not sure if we should try to find a solution for it 🤔

elzadj (approved):

Naaaaiix (approved):
Cool @Marko64 ! 👍 

Update with master to fix the tests and will be ready to merge 👍

PR #2855, opened 2022-04-13
Issue-2810: Replace margin-top on Columns for gap

bilouStrike:
Fixes: #2810


Marko64 (approved):

Naaaaiix (commented):
Some things to consider:
1. `Remove gap` toggle on Row Maxi was designed to reduce the column width (%) as the space was created whit the `justify-content: space-between`. So, a 5 equal column pattern with gap created 5 columns with a 'general' column width of `18%` (2% is for the gap) and in case gap was disabled it creates 5 columns with a 'general' column width of `20%`. This phenomena is not necessary anymore, as we can set a width of `20%` and the gap will do the rest. That way will be much more dynamic and organic.
2. `Remove gap` toggle is not necessary anymore: set gap 0 and done! lol
3. No need for a new default file for gap => we've got the `flex` default options, can work from there.
4. Need a new component for gap that should be able to deal with `gap` (affecting column and row) or leave creators to chose between `gap-column` and `gap-row`. Kind of AxisControl with the different `sync`

I'll have a deeper look after implementing these ideas 😋

FitseTLT (approved):

PR #2856, opened 2022-04-13
Issue 2832 - slider mobile touch actions

myroslavve:
Fixes #2832 
Also made slider's translate reset when it is moved less than 100px, so it doesn't leave where it was on the end of drag.


Marko64 (changes requested):
That fixed it.  
But, the last slide gets stuck:

Screenshot_2

Naaaaiix (approved):
Perfect 👍

Marko64 (approved):
Works good now actually :) Good job.

bilouStrike (approved):

PR #2863, opened 2022-04-14
Issue 2405 - Add inlineStyles to improve color picker and transform using experience

Fybex:
Fixes #2405 

Changes:
- in withMaxiProps HOC two new functions was created - insertInlineStyles, cleanInlineStyles which are using in inspector-tabs, inspectors, toolbar to add inline styles to make maxi smoother in color picker or transform.
Example: 

https://user-images.githubusercontent.com/46112160/163389181-99b85c5e-20df-4f03-9a82-85d928dcca7a.mp4

- editDispatch which was used to change svg content not HOC now and all functions from him was moved to extensions/svg
- AccordionItemPanel not rendering when isHidden which increase optimization

How to test: change colors and transform options in all blocks from sidebar and toolbar to check if it works correct(smooth color picker, no errors in console, inline styles are removing after change ending, etc)

Thanks to @Naaaaiix to his first implementation, which helped a lot in developing process 👍🙂


Marko64 (approved):
Muuch faster.

elzadj (approved):

bilouStrike (approved):

Marko64 (approved):

Naaaaiix (commented):
Is an incredible and beautiful implementation, the results are awesome 😸 good job @Fybex 🥳 🎉 

Some small improvements:
1. Transform `insertInlineStyles` parameters into a single object: the reason is that this function will be constantly growing (onces this branch will be merged, will need to adapt it for AdvancedNumberControl also), so we might have new parameters in future. Having to write all parameters when we are not going to use them can be tedious, an object is much easier to read and to work with 👍 
2. Abstract `insertInlineStyles`, `modifyStyleElement` and `cleanInlineStyles` into the same folder of `src/extensions/maxi-block`, even in the same file (as they are part of samefunctionality) in order to leave `withMaxiProps.js` as clean as possible. If not that file can start growing without end 👍

Naaaaiix (approved):
Awesome 👏 🙌

FitseTLT (approved):

PR #2868, opened 2022-04-15
Issue 2459 Added Responsive Tabs for Sidebar

FitseTLT:
fixes #2459


Naaaaiix (changes requested):
> For me Border hover works fine but if there is specific block on which border hover is not working, please let me know.

It works good for me also, but not sure if we are confusing situations: this issue is to implement the responsive bar with the taps, is possible indicators doesn't work but this is not the thread 👍  @Marko64 what looks broken for you?

> SVG colors: Yes, I have removed responsive tabs because there is no breakpoint variations for SVG colors for now.

Will be implemented soon; as long as is not implemented yet I consider a correct decision to don't add it 👍 

> Number color: You are right but inside NumberCounterControl there is number counter width which has breakpoint variation but other options have no breakpoint variations. So what do you think @Naaaaiix Should I make other options disappear when not in general? Because even if I apply the responsive tabs for the width prop only it would still be confusing b/c NC width is the first prop in the NCC component.

That is kind of a general law we've been all implementing without saying: if the setting doesn't have responsive attributes, just show in general 👍 

About tests: on background we need to fix these lines, as they are changing the responsive stage:
https://github.com/yeahcan/maxi-blocks/blob/d0f61c525259ade1876318b2619cbc45d65442d6/e2e-tests/utils/addBackgroundLayer.js#L4-L9

bilouStrike (approved):

PR #2877, opened 2022-04-19
Svg background test

Alejandro9kgm:
With this test, we prevent this problem https://github.com/yeahcan/maxi-blocks/pull/2798#issuecomment-1098902137 from recurring.


Naaaaiix (commented):
Test works great, but I don't understand how the test is going to prevent having [this issue](https://github.com/yeahcan/maxi-blocks/pull/2798#issuecomment-1098902137) again 🤔

Naaaaiix (commented):
Better. Couple of things:
1. Avvoid `$$` and replace for stronger css classes. That helps a lot on future changes 👍 
2. Add a `(await getEditedPostContent()).toMatchSnapshot();` before disabling hover background to have a snapshot of the styles with the hover 👍

Naaaaiix (approved):
> In this specific test these cases I don't think the $$ are too much of a problem, as they are quite specific, except for the ones on lines 40 and 67 but as they are colours selected from palettes, or the one on line 36 even if it has the $$ it uses specific classes.🤔

One of the main problems we find when updating tests are the $$ selectors. If you consider they are correct like this let me know and I'll merge the test 👍

Naaaaiix (approved):
Much cleaner, bravo 👏  thanks @Alejandro9kgm 😸

PR #2885, opened 2022-04-20
Issues 2768 Combined parentBlockStyle and blockStyle

Fybex:
Fixes #2768

parentBlockStyle and blockStyle was combined into one variable - blockStyle which has two possible values: dark and light


Naaaaiix (commented):
Good job 👍 Looks much more clean and we'll avoid confusions between parentBlockStyle, rawBlockStyle, blockStyle and so on... lol

I think we can 🔥 `defaultBlockStyle` attribute also 👍  what do you think?

About tests: please, don't do a bulk update of snapshots. Most of them are returning error due to the remove of `parentBlockStyle` attribute, but please, check one by one. I know is tedious but we don't know the scope of the implementation and they may show us something is broken 🙏 😸

Alejandro9kgm (approved):

Naaaaiix (commented):
Check the tests, something runs bad on them 🤔

Naaaaiix (approved):
Really good job @Fybex, thanks 😸

FitseTLT (approved):

PR #2909, opened 2022-04-22
Issue 2796 - Add row border-radius to inner columns

Fybex:
Fixes #2796 

Create new attribute to column-maxi - parentBorderRadius which take borderRadius styles from row and apply it to left and right columns if they haven't their own.

Also, what should we do if row will have padding? Remove border-radius to inner columns?
image


Marko64 (approved):
I think we can leave the column border radius as is, even when the row has padding.
You already made it so that you can override column border radius by adding it manually, so that should be enough.

bilouStrike (approved):

Naaaaiix (commented):
Good 👌 

One last thing: please, update `src/extensions/styles/helpers/test/getBackgroundStyles.js` unit test and generate unit tests for `src/extensions/styles/getParentBorderRadius.js`

Naaaaiix (approved):
Perfect 👍 😍

elzadj (approved):

PR #2950, opened 2022-04-28
Issue 2949 - Padding reload bug

Fybex:
# Description
Padding wasn't working after saving and reloading because its attribute was 'number', not 'string' as the margin has.

Changes:
- change padding attributes type to 'string'
- change button-maxi, container-maxi paddings to string

Fixes #2949 

# How Has This Been Tested?
I was running it with group, container and button to check if padding works on responsive, button presets after reload.

# Test checklist

**_ Front/Back Testing _**

**_ Pre-Code Testing _**

# Checklist


Marko64 (approved):

FitseTLT (approved):

myroslavve (approved):

Naaaaiix (commented):
Well, let's start a controversial discussion lol

The fix works, but is transforming `padding` attributes to 'string'. Is the same happened with a @FitseTLT implementation some days ago and [here is my answer](https://github.com/yeahcan/maxi-blocks/pull/2911#discussion_r857736901). With this current implementation we face:
1. Unexpected errors: need to check deeply the helper, the possible side-effects.
2. Library: we need to ensure the current attributes on the created patterns of the library doesn't break.

The problem comes from this line:
https://github.com/yeahcan/maxi-blocks/blob/3083338d7d81902541875d709d0777d598b94058/src/components/inspector-tabs/inspector-margin-padding.js#L81

So, we have to take a decision:
1. Transform to 'string' as it is, checking every detail related with the padding group attributes and we don't break any Library pattern.
2. Find another solution for the line shared before.

And this suggests me another question:
Should we swap all number attributes like `padding`, `width`, `height` and much others from `number` to `string`? And is not a shallow question: we saw what happen when we change attributes and we've got some job done, it happened with the new grid implementation. Now what we risk is the Library patterns we've got, but onces Maxi will be released we won't have that facility. 

Please @Fybex confirm us what happen with created patterns when passing from master to this branch, lets hope a miracle makes we don't lose the values set 👍

elzadj (approved):

PR #2974, opened 2022-04-29
issue-2937-button-text-select-error

Marko64:
# Description

Problem: _Text direction_ is causing an error. It needs a default value.
I'm not sure if this is the correct way to fix it. If not, please suggest where else can I define the defaults.

Fixes #2937

# How Has This Been Tested?

Insert the text or the button block and select it. Check the console for errors. 

# Test checklist

**_ Front/Back Testing _**

-   [ ] Inset the text block. Select it. Check the console for errors.
-   [ ] Select the Typography panel from the sidebar and test all the settings that use SelectControl. 
-   [ ] Repeat the same for the button block.

**_ Pre-Code Testing _**

-   [x] Inset the text block. Select it. Check the console for errors.
-   [x] Select the Typography panel from the sidebar and test all the settings that use SelectControl. Make sure they work and don't throw any errors in the console.
-   [x] Repeat the same for the button block.

# Checklist


Alejandro9kgm (approved):

Naaaaiix (commented):
This implementation 'fixes' partly the issue, the problem is that can generate a new one. Adding a default attribute we create the CSS, which can go against the creators interests when is rtl website. A better fix would be, on:
https://github.com/yeahcan/maxi-blocks/blob/e489d3e3d300646b2a55f080a3108946d18f9955/src/components/typography-control/index.js#L876

replace for:
```
			<SelectControl
				label={__('Text direction', 'maxi-blocks')}
				className='maxi-typography-control__direction'
				value={getValue(
					`${prefix}text-direction`,
					breakpoint,
					false,
					true
				)}
[...]
````

Alejandro9kgm (approved):

FitseTLT (approved):

Naaaaiix (changes requested):

Naaaaiix (commented):
Tests need an update 😸

PR #2994, opened 2022-05-03
Add test for column border

Alejandro9kgm:
In this test we check the maxi column border, the normal state and the hover state, the rest like for example the responsive is already tested in border-control.

https://github.com/yeahcan/maxi-blocks/pull/2909


Naaaaiix (changes requested):
This test checks the border-radius set on a column. What has been requested is to check that, when adding border-radius to the Row, the first and last inner columns get the corresponding border-radius depending on its position (left border-radius for first column, right border-radius for last). Please, check the #2909 and #2796 again 👍

Naaaaiix (approved):

PR #3053, opened 2022-05-10
Test Advanced number control limits

Alejandro9kgm:
With these tests we check that the min and max values are working correctly.


Naaaaiix (changes requested):
Is strange we hadn't this test of the min/max before, and is good we've got it now, but this is not what was requested 😅 

#2705 implemented limits on the `range`, and what was requested to do was to test them. Please, enhance the tests you've implemented adding this one 👍

Naaaaiix (approved):
👏

PR #3065, opened 2022-05-11
Issue 3020 - Column height correct work

Fybex:
# Description

Fixes #3020 

# How Has This Been Tested?

Changed height, force aspect ratio state on different responsive breakpoints and checked if all work well with column-maxi

# Test checklist

**_ Front/Back Testing _**

**_ Pre-Code Testing _**

# Checklist

### DIFF
--- a/src/blocks/column-maxi/edit.js
+++ b/src/blocks/column-maxi/edit.js
@@ -86,7 +86,10 @@ class edit extends MaxiBlockComponent {
 			const columnHeightObj = this.getHeightObj();
 
 			if (
-				this.resizableObject.current.state.width !== `${columnWidth}%`
+				this.resizableObject.current.state.width !==
+					`${columnWidth}%` ||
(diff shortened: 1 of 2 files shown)


Marko64 (approved):

myroslavve (approved):

Naaaaiix (commented):
There's a possible side-effect we need to be care of, and is the fact that at the first render Column resizableBlock doesn't have a default height (so is relying on the default maxi `height: auto`), when adding the force aspect-ratio it becomes 100% and then, when disabling the force aspect-ratio it stays there. Not sure if it can generate some unexpected experience tho:


https://user-images.githubusercontent.com/50477222/168234211-af854c2f-e34b-49a5-92ce-26b0c85ca09c.mp4

Naaaaiix (approved):
Gorgeous 🤩

PR #3073, opened 2022-05-12
Fix and update indicator tests Margin/Padding

Alejandro9kgm:
The indicator tests have been fixed and responsive tests have been added.


Naaaaiix (commented):
The test is not requesting all the parts we need to check. What we need to get also is the HTML of this elements:

Captura de pantalla 2022-05-16 a las 18 15 12

Naaaaiix (changes requested):

Naaaaiix (approved):

PR #3140, opened 2022-05-18
Issue 3104 block sidebar header ui

Marko64:
I'm not sure I've added the info icon and the state hook at the right place. If it should be better added somewhere else, let me know.

# Description

Adds more compact style to the block's header in the sidebar.

Fixes #3104


Rustam198804 (approved):

FitseTLT (approved):
Nice!!! I like it!!!

Naaaaiix (approved):
It works good, but as commented, the way `useState` and the way to affect `'.block-editor-block-card__description'` are not the optimal. Have a look to [this file](https://github.com/yeahcan/maxi-blocks/blob/9a223d80e78f99fc88b6ac9e7cc4f2744a52ceb1/src/components/block-styles-control/index.js#L43-L53) to see some suggestions about. If you decide to add them re-request the review, if not, merge directly 👍

elzadj (changes requested):

PR #3148, opened 2022-05-19
Update and fix Button Maxi test

Alejandro9kgm:
The button maxi block test has been fixed and updated.


Naaaaiix (commented):
After testing icon options on normal and hover, check frontend result and then check editor results after reloading backend 👍

Naaaaiix (commented):
Cool! 

With the new updates this test requires need improvements. We need to check the `inherit` as has been commented on #3225 👍 

> Well, this is going to be tricky and long lol
> 
> Test default button presets work correctly
> Test a Button Maxi with icon and its whole settings (including hover)
> Set inherit and check:
> Icon stroke is equal to the typography colour
> Background of icon is equal to button background, and if there's no button background, it becomes transparent
> If typography hover is active, on hover the button, icon stroke is equal to typography hover colour
> if button background hover is active, on hover the button, icon background is equal to background button colour
> With inherit active, enable Icon only. In that case
> Icon stroke colour can be changed from Icon options
> Icon background comes from the button background in case it is there; also the hover.
> Removing inherit and with active icon only, background should be accessible from Icon options
> As you can see, there are different combinations, so please test them all as much as you can. Also consider testing 'filled' icons, as there was a bug that, when setting inherit it wasn't showing fill colour properties 👍
> 
> I'd really appreaciate that, if you find a bug, record a video or comment the steps to arrive there, as this amount of different combinations make it hard to deal with Button Maxi. Time to play 😸

myroslavve (changes requested):

myroslavve (approved):

Naaaaiix (changes requested):
I think it still needs these settings tests for the icon test:
1. Icon position
2. Icon background
3. Icon gradient

The `inherit` test is not testing what happens if the Typography or BG hover is not set, and what happen when the values of that settings are modified. Also the `Icon only` logics: colour can be changed if is enabled, and can't be changed when is disabled. We need to check ALL these logics to ensure we don't break them in future 👍

PR #3188, opened 2022-05-24
Add new toolbar test

Alejandro9kgm:
In this Pr the tests background-color, block-background-color, border, box-shadow, divider-alignment, divider-color, divider-line are added.

In the test what you do is edit the blocks from the toolbar, then check that the attributes have been changed correctly and then check that the changes are also shown in the sidebar.


Naaaaiix (changes requested):

Naaaaiix (approved):
Well done @Alejandro9kgm ! 👏 leaving comments on every action makes the reading and understanding of the test much more easy, thanks. Keep with it 💪

PR #3247, opened 2022-05-31
Copy Past UI

Rustam198804:
# Description

Fixes #3211


Marko64 (approved):

Fybex (changes requested):
When hover on text, which has nested options, hover for checkbox doesn't apply, and text colour different from texts without nested options
copypaste-ui

Fybex (approved):

Naaaaiix (approved):
Looks good, thanks @Rustam198804 😸 

Found a small tiny thing I'll derive into a new issue, but is more related with the logics of the component than the styles, so is not related with this implementation 👍

PR #3258, opened 2022-06-02
Add block resizer test

Alejandro9kgm:
This is the first test that works this way, the test works by generating the number counter block and then click & drag to resize the block.


Naaaaiix (approved):
Astonishing, really. Incredible job @Alejandro9kgm, well done 👏

PR #3279, opened 2022-06-06
Border control responsive test

Alejandro9kgm:
Add responsive test to border control


Naaaaiix (commented):
Cool, it works great. Would be awesome if we can implement another test, like the responsive one, but this time based on clicking on the defaults; most likely we need to use the default button to remove the block on some responsive stage 👍

Naaaaiix (changes requested):

Naaaaiix (approved):

PR #3310, opened 2022-06-13
Issue:3290 NC text alignment option

FitseTLT:
# Description
Added text alignment option for NC

Fixes #3290 

# How Has This Been Tested?
Tested Text alignment option is working for NC

# Test checklist

**_ Front/Back Testing _**

**_ Pre-Code Testing _**

# Checklist


Marko64 (approved):

Fybex (approved):
All works, but some minor things to consider:
1. Should we show `textAlign` even when circle is showed? 🤔 
2. Should `textAlign` have `center` by default?

Fybex (changes requested):
image

Marko64 (approved):

Fybex (changes requested):
All good,  just delete code from previous implementation with one more alignment and:

Fybex (approved):

Naaaaiix (changes requested):
The idea is good, but IMO is overcomplicated and can be confusing for creators. I think it will be easier if we:
1. Remove the new toggle for `Set widht to fit content`
2. When removing the circle it directly disappears width options on `Number` accordion
3. Instead of using `fit-content`, use `auto` which has better compatibility between browsers
4. Instead of using CSS to overwrite BlockResizer styles using `!important` use its API method `updateSize`. You can find different examples on our blocks, like Column Maxi 👍

Naaaaiix (changes requested):
Tests results are quite strange; they seem to share some error. Try to trace them please 👍

Naaaaiix (changes requested):
Check it deeper, e2e tests shows something is broken 👍

Naaaaiix (changes requested):
Still finding some small issues 👍 

https://user-images.githubusercontent.com/50477222/176224852-a95261c8-dda1-4b55-b2e6-ef155c675bf2.mp4

Marko64 (approved):

Fybex (approved):
Not sure, but maybe on auto we should just remove resizing as we are doing in image-maxi with BlockResizer showHandle prop 🤔  Just thoughts 👍

Naaaaiix (commented):
Much better, thanks @FitseTLT 😸 🤟 

I agree with @Fybex, in order to keep the coherence with Image Maxi and in what is IMO a better UX, when auto is enabled the resizable handles should disappear 👍 the use of BlockResizer showHandle prop will help on it

Naaaaiix (approved):

PR #3314, opened 2022-06-13
Issue:3293 fixed shape scale bug for bg layers

FitseTLT:
# Description
fixed shape scale bug for bg layers

Fixes #3293 

# How Has This Been Tested?
Added shape bg layer set use bg image and then click down arrow key to decrease shape scale.

# Test checklist

**_ Front/Back Testing _**

-   [x] Test the component in hover state in sidebar (if hover exists)

**_ Pre-Code Testing _**

-   [x] Test the component in hover state in sidebar (if hover exists)

# Checklist


Marko64 (approved):
Fixed. 

Noticed that shape doesn't work for containers. Should I make a new issue for that, or you might want to fix it here?
Screenshot_4

myroslavve (changes requested):

myroslavve (approved):

Naaaaiix (changes requested):
With the new implementation I'm finding many `transform: scale(1) ;` created with the `getImageShapeStyles`. I think in case we are having it, we shouldn't display, as it doesn't have much sense, and we can reduce the amount of CSS 👍

Naaaaiix (changes requested):
Still having this thing with the unnecessary `scale(1)` and problems with the responsive 👍

Naaaaiix (approved):

PR #3379, opened 2022-06-23
Issue 3372 - Overflow responsive default fix

Fybex:
# Description

All html elements have overflow visible by default, so I've decided to remove default option from options, because it hasn't sense and not working on responsive. Now we have overflow visible by default.

Changes:
- removed default option
- set overflow attributes visible by default
- ensured that we are not adding unnecessary css (because of setting visible overflow as default)
- added unit tests

Fixes #3372 

# How Has This Been Tested?

# Test checklist

**_ Front/Back Testing _**

**_ Pre-Code Testing _**

# Checklist

### DIFF
--- a/src/components/overflow-control/index.js
+++ b/src/components/overflow-control/index.js
@@ -33,7 +33,6 @@ const OverflowControl = props => {
 					label={__(`Overflow ${axis}`, 'maxi-blocks')}
 					key={uniqueId('maxi-position-control__overflow')}
 					options={[
-						{ label: 'Default', value: '' },
 						{ label: 'Visible', value: 'visible' },
 						{ label: 'Hidden', value: 'hidden' },
 						{ label: 'Auto', value: 'auto' },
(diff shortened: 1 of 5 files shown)


Marko64 (approved):

FitseTLT (changes requested):
What about following the idea I used here?
https://github.com/yeahcan/maxi-blocks/blob/dd6731615f7026d8f887185b5285490be8ee4857/src/extensions/styles/helpers/getImageShapeStyles.js#L21-L55
It will be more effective then no need for another loop to check if the style is needed and also  for instance M breakpoint style will not affect L screen sizes so no need to add style for L if only M overflow is changed and so on.

FitseTLT (changes requested):

FitseTLT (approved):

elzadj (changes requested):
ignore, was some sort of a weird local glitch 

=============================


All is good, but I think we need to add 
clip;
scroll;

to the options too.
Or maybe just all possible values https://developer.mozilla.org/en-US/docs/Web/CSS/overflow#syntax

elzadj (approved):

PR #3406, opened 2022-06-28
Add the video maxi block and video-control tests.

Alejandro9kgm:
In this Pr the first tests for Video Maxi block, and the test for video-control are added.


FitseTLT (approved):

Naaaaiix (approved):
The test is good, just I found this situation which I'm not sure is correct. Please @Alejandro9kgm test it, and create an issue if is not working as it should 👍

PR #3407, opened 2022-06-28
Add video-options-control test

Alejandro9kgm:
In this pr the tests for video-options-control are added


FitseTLT (approved):

Naaaaiix (commented):
Most of the attributes modified on this component need to be checked on the frontend part and aren't styles, so the snapshot is not apporting much. To really test this case, is necessary to check the HTML frontend result 👍

Naaaaiix (changes requested):
Tests are broken 👍

PR #3411, opened 2022-06-29
Issue 3381 - Blinking transitions

Fybex:
# Description

Added `-webkit-transform: translate3d(0,0,0)` on image layer opacity

Fixes #3381 

# How Has This Been Tested?
Tested image layer functions on this [code_editor](https://github.com/yeahcan/maxi-blocks/files/9007668/blinking.txt), if they work as before, but without blinking on hover

# Test checklist

**_ Front/Back Testing _**

-   [x] Test if we haven't blinking any more
-   [x] Test if other image layers functions works as they were

**_ Pre-Code Testing _**

-   [ ] Test if we haven't blinking any more
-   [ ] Test if other image layers functions works as they were

# Checklist

### DIFF
--- a/src/extensions/styles/helpers/getBackgroundStyles.js
+++ b/src/extensions/styles/helpers/getBackgroundStyles.js
@@ -414,7 +414,10 @@ export const getImageBackgroundObject = ({
 	// Opacity
 	if (isNumber(bgImageOpacity)) {
 		response[breakpoint].opacity = bgImageOpacity;
-		if (!isHover) {
+
+		const hoverStatus = props[`${prefix}background-hover-status`];
+
+		if (hoverStatus) {
 			// To avoid image blinking when hovering
 			response[breakpoint]['-webkit-transform'] = 'translate3d(0,0,0)';
 		}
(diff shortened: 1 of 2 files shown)


Marko64 (approved):
I think this is worth re-checking one day when we have more time to deal with it. It is fine for now.

FitseTLT (changes requested):
Still happening for me in frontend.
https://user-images.githubusercontent.com/38649957/176410965-cf962e58-a3fa-4d83-8a3b-c36d555a5b64.mp4

FitseTLT (approved):

Naaaaiix (commented):
Going to say something that maybe is ridiculously simple and I didn't see it, but... why don't we add this `-webkit-transform: translate3d(0,0,0);` property directly in the static CSS instead of creating it dynamically when creating the styles? Don't kill me if it's so obvious lol 

In other order of things: yesterday on the meeting I said I remember some article about this glitch and found it. Here it talks about [this flickering error](https://maximelafarie.com/avoid-css-flickering) that I couldn't reproduce (on the website shared they say that if you hover on the images you can experience it: I can't! lol)
As I've been reading, is related with the computer performance. CSS property like `will-change` are used to prepare/notify to the hardware that a change is coming so need can optimize the resource for that action. [This stackoverflow answer](https://stackoverflow.com/a/43046569) talks about other options, and I think we need to think in terms of performance in this concrete case, as seems a problem with [GPU and WebGPU pergormance](https://bugs.chromium.org/p/chromium/issues/detail?id=961777&q=flick%20transition&can=2). 👍

PR #3415, opened 2022-06-30
Fix Group link inside a column

elzadj:
# Description

Fixes #3383 

Was careful to not affect all .maxi-link-wrapper, so added it for links directly inside columns only (for now). Not sure if we need it something else. :thinking: 

# How Has This Been Tested?

Test groups with links inside columns. The height must be equal. 

Compare before:

Local WordPress (4)

And after:

Local WordPress (5)

[code.txt](https://github.com/yeahcan/maxi-blocks/files/9018872/code.txt)

# Test checklist

**_ Front/Back Testing _**

**_ Pre-Code Testing _**

# Checklist

### DIFF
--- a/src/blocks/column-maxi/style.scss
+++ b/src/blocks/column-maxi/style.scss
@@ -5,4 +5,8 @@ body.maxi-blocks--active div.maxi-column-block {
 	height: auto;
 	margin: 0;
 	box-sizing: border-box;
+
+	> .maxi-link-wrapper {
+		height: 100%;
+	}
 }


Marko64 (commented):
That works. A couple of notes:
- If a column has a background layer, the link wrapper has a lower z-index. Maybe we should make the default z-index for links higher.
- Checked adding links to columns directly, and the wrapper also needs 100% height, and 100% width too.

Marko64 (approved):
Yes, that would be better and def more logical. Let's do it.

Olekrut (approved):

myroslavve (approved):

Naaaaiix (commented):
Not sure about the `height:100%` and the possible side effects it can have. Here's a video where I show one of the possible side effects, and a possible alternative to fix this issue. Not sure about my suggestion also, so let me know your thoughts about 👍 

https://user-images.githubusercontent.com/50477222/176699534-18e326db-f4d4-4a11-89ac-9feec891a8f2.mp4

PR #3477, opened 2022-07-13
Loading indicator for local fonts #2854

### LINKED ISSUE #2854: Loading indicator for local fonts
For the dashboard, on post save and on SC apply / save.
Something like "Downloading fonts..." and a loading bar. Want it to be dynamic so people do not think it's stuck.

>  what ever is easiest and just make it orange

https://codepen.io/Hadeon/pen/jMNbay
https://codepen.io/danhearn/pen/MvqgdM
https://codepen.io/Last_dreamer/pen/XWKMzXW
https://codepen.io/egecavusoglu/pen/povjGEr

FitseTLT:
# Description
implemented font download notice

Fixes #2854

# How Has This Been Tested?
Check font downloading notices both in success and fail state.

# Test checklist

**_ Front/Back Testing _**

-   [x] Check font downloading notices both in success and fail state.

**_ Pre-Code Testing _**

-   [x] Check font downloading notices both in success and fail state.

# Checklist

### DIFF
--- a/src/extensions/text/fonts/loadFonts.js
+++ b/src/extensions/text/fonts/loadFonts.js
@@ -62,34 +62,50 @@ const loadFonts = (font, backendOnly = true, target = document) => {
 					`url(${url})`,
 					fontDataNew
 				);
-
-				dispatch('core/notices').createNotice(
-					'info',
-					__('Downloading font…', 'maxi-blocks'),
-					{
-						id: url,
-						type: 'snackbar',
-					}
-				);
+				let fontLoaded = false;
+				let noticeCreated = false;
+				let downloadFailed = false;
+				if (target.fonts.check(`12px ${fontName}`)) fontLoaded = true;
+				else {
+					setTimeout(() => {
+						if (downloadFailed) return;
+						if (target.fonts.check(`12px ${fontName}`))
+							fontLoaded = true;
+						else {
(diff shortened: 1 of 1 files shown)


Marko64 (commented):
I'm not sure what we are getting with this.
- If you select the same font multiple times it will always say "font downloaded". Is that correct?
- As soon as you open SC, it says "font downloaded".
- When you change the font in the SC manager, it has "font downloaded" 2 times.

Thinking out loud, people may be worried if they are downloading something unnecessary and storing it somewhere on the server.

How can I recreate the original issue for which we decided this notice would come in handy?

Marko64 (approved):
I think that's better, thanks.

myroslavve (approved):

elzadj (changes requested):
I don't think the loading indicator works at all. 

When I save a page, it updates right away, and then the frontend fonts are broken for some time, because maxi still downloads them. And I need to w8 and hit f5 several times too see the new fonts.

Ideally the indicator should be shown after user hits Update. And the "Page Updated" messages should be on hold until all the fonts are ready.
[video-10085862-8b4c9aedabf63604dd505a2a7ef0df53.webm](https://user-images.githubusercontent.com/2401618/180207660-d4393c43-9a49-4619-8727-559657a350ea.webm)

PR #3484, opened 2022-07-15
Text selected when changing settings in the toolbar #3432

FitseTLT:
# Description
fixed the text selection issue on changing prop from toolbar
fixes on the memory leak warning of the toolbar popover

Fixes #3432 

# How Has This Been Tested?
tested the text is not selected on changing settings from toolbar

# Test checklist

**_ Front/Back Testing _**

**_ Pre-Code Testing _**

-   [x] Same 1-4 points for the toolbar

# Checklist

### DIFF
--- a/src/components/toolbar/captionToolbar.js
+++ b/src/components/toolbar/captionToolbar.js
@@ -125,7 +125,11 @@ const CaptionToolbar = memo(
 				returnFormatValue: true,
 			});
 
-			const newFormatValue = { ...obj.formatValue };
+			const newFormatValue = {
+				...obj.formatValue,
+				start: formatValue.start,
+				end: formatValue.end,
+			};
 			delete obj.formatValue;
 
 			onChangeTextFormat(newFormatValue);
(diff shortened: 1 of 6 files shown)


Marko64 (approved):

myroslavve (changes requested):
Found same thing in caption toolbar of image.

https://user-images.githubusercontent.com/53875005/180755217-271a384c-517a-4cb4-9b5c-809a06dedd0f.mp4

I believe the problem is also with `newFormatValue`, but in this file:
`src\components\toolbar\captionToolbar.js`

myroslavve (approved):

Naaaaiix (commented):
The fix is correct, but maybe there's a higher optimization. I mean, all the code implemented is a modification of the FormatValue retuned from `setFormat`. Could it be better to modify the function to return the original `start` and `end` instead of modifying the returned object every time we request it? 🤔

Naaaaiix (approved):

PR #3543, opened 2022-08-04
New component - Text Input

Marko64:
*New component: Text Input

Fixes #3535

# What to expect

Now, changing the block label or the custom css class on the text, shouldn't get slow 👍


myroslavve (changes requested):

myroslavve (approved):

Fybex (changes requested):
Styles aren't cleaned, input has borders, input size isn't comes from content size

Fybex (changes requested):

Fybex (approved):

Naaaaiix (changes requested):
The new component looks awesome, thanks @Marko64. What is also necessary is to add it inside TextControl: the input element should be replaced with this new component 👍

Naaaaiix (commented):
Tested, and is returning an error:


https://user-images.githubusercontent.com/50477222/191529171-819ec76a-4e01-4f0d-8fbb-dda8114401c0.mp4

I've written more comments, but not sure why they aren't appearing. Well, basically:
1. The styles are broken cause on TextInput, the `className` prop is not used
2. Not sure about `valueChange` function and if it's going to work as expected. I recommend you reading Text Maxi block `edit.js` to see the timeout is a variable and is 'cleared' if the next action is called before the timeout function is called 👍

myroslavve (approved):

Fybex (changes requested):

Olekrut (approved):

Fybex (approved):

elzadj (approved):
Much, much faster, thank you.

I think we need to deal with the spaces for the 'classes' field. Maybe add a placeholder "Separate multiple classes with spaces", or something like that? Need to gently remind people that we can't have spaces inside a single class name:  

screenshot-nimbusweb me-2022 11 01-17_47_56

What do you think @Marko64 @ckp267 ?

PR #3644, opened 2022-09-01
Issue 3636 - fix values not saving on unit switch

myroslavve:
# Description
Didn't work because even with `optionType === 'string'` it used numbers in `onChangevalue`.

Fixes #3636 

# How Has This Been Tested?
Checked that values that were set to not exceed the max value are saved after reload.

# Test checklist

**_ Front/Back Testing _**

-   [x] Check that values that were set to not exceed the max value are saved after reload

**_ Pre-Code Testing _**

-   [x] Check that values that were set to not exceed the max value are saved after reload

# Checklist

### DIFF
--- a/src/components/advanced-number-control/index.js
+++ b/src/components/advanced-number-control/index.js
@@ -224,7 +224,11 @@ const AdvancedNumberControl = props => {
 
 								if (value > minMaxSettings[val]?.max) {
 									onChangeValue(
-										minMaxSettings[val]?.max,
+										optionType === 'string'
+											? minMaxSettings[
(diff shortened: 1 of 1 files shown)


Marko64 (changes requested):
This fixes the issue, but we also need to increase the max value for the % unit. Normally texts have somewhere between 100 and 200 line height when % is used. Let's change the limit to 200.

Marko64 (approved):

Fybex (approved):

elzadj (approved):

PR #3744, opened 2022-09-29
Fix 3728 - Set Custom CSS to display selected image

blockchain-web-mobile-dev:
I have set custom css as "display: block" to display the selected image.

Fixes #3728 

# How Has This Been Tested?

# Test checklist

**_ Front/Back Testing _**

-   [ ] Check the Solution is good for issue

**_ Pre-Code Testing _**

# Checklist


Marko64 (commented):
If I understand the issue, the problem is that when you select the image, you expect to be able to re-upload it as you would normally do; however, this image uses a background image as well so it's not clear that you need to go to canvas settings to change it there.

This solution doesn't seem to help with the issue, and I'm not a fan of forcing a CSS on an element by default.

PR #3757, opened 2022-09-30
Fix 3634 - Fit column after use column gap

blockchain-web-mobile-dev:
I have set initial value column-fit-content value each breakpoints(xxl, xl, etc...).

https://watch.screencastify.com/v/CBPGLrnsXmYArBSnTGrg
please check this video.
If user didn't set the column-fit-content for each breakpoints, the column gap will not work.

Fixes #3634 

# How Has This Been Tested?

# Test checklist

**_ Front/Back Testing _**

-   [ ] Check this is good solution for this issue

**_ Pre-Code Testing _**

# Checklist

### DIFF
--- a/src/components/column-size-control/index.js
+++ b/src/components/column-size-control/index.js
@@ -17,10 +17,33 @@ import SelectControl from '../select-control';
 import { ToggleSwitch } from '../../components';
 
 /**
+ * General
+ */
+const breakpoints = ['general', 'xxl', 'xl', 'l', 'm', 's', 'xs'];
+
+/**
  * Component
  */
 const ColumnSizeControl = props => {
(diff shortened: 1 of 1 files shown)


Naaaaiix (changes requested):
As commented on the same issue #3634, the problem is related with BlockResizer component. On changing the flex gap it gets back the old `width` and `flex-basis` values, and doesn't keep the `fit-content` ones:


https://user-images.githubusercontent.com/50477222/193523460-6157c588-20c7-4c4d-a0c2-f09224bc1db1.mp4

The solution proposed on this PR isn't correct: we can't be changing all the responsive attributes every time we change the button; that means nullifying the responsiveness we've got 🤔

PR #3712, opened 2022-09-23
Issue 3700 - Remove hoverProp from transition attributes

Fybex:
# Description

Update:
- removed `hoverProp` from transition attributes
- `hoverProp` key now is always required in transition data in data.js of maxi-block
- `limitless: true` → `property: false`
- `ignoreHoverProp: true` → `hoverProp: undefined`
- other small fixes

Fixes #3700 

# How Has This Been Tested?
Checked if all transitions still work, checked if svg background transitions work in [this code editor](https://github.com/maxi-blocks/maxi-blocks/files/9635527/svg-with-old-hover-prop.txt).

# Test checklist

**_ Front/Back Testing _**

-   [x] Test transitions activation when hover state is turned on, deactivation when hover is disabled
-   [x] Test transitions work as usual

**_ Pre-Code Testing _**

-   [ ] Test transitions activation when hover state is turned on, deactivation when hover is disabled
-   [ ] Test transitions work as usual
-   [ ] Check no commented code and no unnecessary imports
-   [ ] Standards of the project have been followed
-   [ ] No errors/warnings on console

# Checklist

-   [X] My code follows the style guidelines of this project
-   [X] I have performed a self-review of my code
-   [X] My changes generate no new warnings/errors
-   [ ] I have added/updated tests that prove my fix is effective or that my feature works
-   [X] New and existing unit tests pass locally with my changes

PR #3777, opened 2022-10-05
Issue 3672 - Add mechanism to update transition attributes on change transition obj

Fybex:
# Description

Changes:
- created `SVGTransitionMigrator` which checks `transtion.block` svg-icon-maxi attributes and modify them if it's needed
- renamed `transitionMigrator` to `buttonIconTransitionMigrator`
- created `getTransitionData` helper, which returns `data.transition` by block name
- fixed problem with tabs in transitions (always had only one setting - canvas or block after #3712)
 

Fixes #3672 

# How Has This Been Tested?
Checked if there was no transition in master with this [code editor](https://github.com/maxi-blocks/maxi-blocks/files/9717901/svg-with-old-transition.txt) didn't have transition, checked if the svg colour transition works in this branch. Checked if transition tabs working correct (block and canvas tabs, which appears in blocks with block and canvas settings on turning off change all switch)

# Test checklist

**_ Front/Back Testing _**

-   [ ] Test the code editor (described how above in "How Has This Been Tested?")
-   [x] Test transition tabs correct work

**_ Pre-Code Testing _**

-   [ ] Test the code editor (described how above in "How Has This Been Tested?")
-   [ ] Test transition tabs correct work

# Checklist


Naaaaiix (commented):
Let's see if I understood correctly what has been implemented here 😸 

When rendering a block, when `isMounted`, we iterate over transitions and ensure their structure is correct comparing with the default element, right? The idea is correct, but opens some thoughts I'd like to share:
1. This seems more a `migrator` than a method inside the block. The reason is, on one side organization (seems more a migrator IMO), and the other the capacity to be under a `isElegible` condition that may reduce the time of rendering the block. This implementation will iterate on an object, so 100 blocks will do the same when downloading from Library, and it can (and probably will) slow down the speed of rendering. We might ensure we reduce it using `isElegible`.
2. This kind of `checker` (this implementation checks the correct structure of `transitions` object attribute) could be applied for `relations` or even `background-layers` or `transform`. So, I'm not sure if it's a good idea implementing something like this branch propose, as is an enormous amount of process that for the moment we are applying just for `transitions`, but the same can be done for other `object` attributes and that can be a painful point for our speed. So, maybe we need to just fix this issue as a new migrator focused on adding the SVG transition, and don't check `transitions` structure every time we render a block.

Basically, I'm afraid of the speed reducing effect that this implementation can generate, so maybe we need to simplify and reduce as a simple migrator 🤔

Naaaaiix (commented):
Awesome! 👏 

Need to be adapted to the new blockMigrator, but the essence is working perfect 👍

Olekrut (approved):

Naaaaiix (approved):

myroslavve (approved):

moroz-art-dev (approved):

PR #3754, opened 2022-09-30
Optimization - Reduce migrators combinations to one

Naaaaiix:
# Description

As we have been experiencing, e2e github tests are running on errors, basically `timeout` errors. After checking a bit I've discovered it was taking 30s to create a post and add a block, which is insane if we compare with the same test 7 months ago where it took 5s. With that started to check and saw how migrators combination is delaying the load of Gutenberg an average of 15-20s.

To fix it, the plan has been reducing the combinations that we've created on #3311 to just one combination that deals with all the migrators at same time. So, instead of creating all combinations, we are creating just one. 

Here's a video explaining the problems we were facing and the solution generated:

https://user-images.githubusercontent.com/50477222/193297408-9aaca8ed-8313-4344-b088-461acfd2ce42.mp4

# How Has This Been Tested?

As commented on the video, I've used 3 different code editors of different PRs to check they work on the same code editor. The different PRs and parts are:
1. #3494: from where I extracted a whole column. Check the PR or the video to see what to expect.
2. #3547: from where I took come `relations` pasted on the small Group Maxi (check video to see what to expect)
3. #3638: from where I took `block-background-hover-status` and added to the main Group Maxi

Here's the [code editor](https://github.com/maxi-blocks/maxi-blocks/files/9685543/mixofmigrators.txt); with it, to test this branch, we need to:
1. Check Gutenberg loads faster
2. On pasting the code editor all works as expected and commented on the video
3. (not a review) All tests passes faster than the current average.

Time to play! 👾 

# Test checklist

**_ Front/Back Testing _**

-   [ ] Test the component in normal state in sidebar
-   [ ] Test the component in hover state in sidebar (if hover exists)
-   [ ] Check that the settings work on frontend
-   [ ] Check that backend works and saves the settings after the editor reload
-   [ ] Same 1-4 points for the toolbar
-   [ ] Same 1-4 points on responsive
-   [ ] Test the block inside the grid (Container + Row + Column)
-   [ ] Test the block as a standalone block
-   [ ] Duplicate the block, test that the settings of the first do not affect the second
-   [ ] Test with 2 blocks of the same type

**_ Pre-Code Testing _**

-   [ ] Test the component in normal state in sidebar
-   [ ] Test the component in hover state in sidebar (if hover exists)
-   [ ] Check that the settings work on frontend
-   [ ] Check that backend works and saves the settings after the editor reload
-   [ ] Same 1-4 points for the toolbar
-   [ ] Same 1-4 points on responsive
[...]

PR #3781, opened 2022-10-07
Update - Fix blockMigrator

Naaaaiix:
# Description

[⚠️ Don't need to read this; is an explanation of the process `blockMigrator` has been living and how it ended working as it's working. More a diary or some notes for future devs than anything]

[Also includes #3782]

In order to perform the update of devdemo I've been checking the effects of updating its code editors from patterns. Found that #3754 is fixing some aspects like speed, but breaking some patterns as the migrators aren't working correctly. Basically the problem with the migrators is the fact that Gutenberg doesn't allow `chain update`, that's why at the beginning we decided to create a system that combines all migrators to cover all situation. The result was problematic, as it was shown on #3754, and that amount of combinations ended on a huge slow-down of speed on Maxi blocks, so that PR tried to reduce everything to just one migrator, simplifying. The idea seemed good and Maxi was loading faster, but ended bad when trying to update devdemo; the result has been some broken patterns. The methodology Gutenberg follows can be read on [this file](https://github.com/WordPress/gutenberg/blob/05381d2074f6cfd9d1b6d0f779943291203b829d/packages/blocks/src/api/parser/apply-block-deprecated-versions.js), and what is happening is that blocks iterate the deprecated array looking for an object that passes the validation for the blocks. The problem is, if we have 5 deprecations and 2 passes the validation, Gutenberg will accept the last one, and the previous accepted one is not applied; that's where the forbidden `chain update` affects us, because a block that need more than one update is going to be updated just by one. 
As creating an array with all the possible combinations of migrators was the way that responds to Gutenberg no 'chain updates' requirements, but at the same time is heavy baggage that affects Maxi speed, the solution has been breaking this 'chain updates' law Gutenberg imposed creating a combination of store methods inside 'maxiBlocks' store that save the changes done in the block while running the migrator. So, let's have an example we've got 6 migrators and when pasting a block, it needs 3 of them: in normal Gutenberg system just the last one would be applied, but the iteration has run the 3 `migrate` function. To avoid it and have the 3 migrators applied, we are saving the result of `migrate` function into the store and then, on the next `migrate`, we claim for that results to be used by the `migrate` function. That way we do `chain` updates and without losing speed performance.

# How Has This Been Tested?

Get this [code editor](https://github.com/maxi-blocks/maxi-blocks/files/9732827/footer.txt) and check it on `master`, icons position should be broken. Then test the same on this branch and should work correctly. To ensure it doesn't break what was done on #3754, check this other [code editor](https://github.com/maxi-blocks/maxi-blocks/files/9731980/mixofmigrators.txt) that was used there and see it acts as required on that PRs test.

# Test checklist

**_ Front/Back Testing _**

-   [ ] Reproduce the commented test

**_ Pre-Code Testing _**

-   [ ] Reproduce the commented test

# Checklist


Olekrut (approved):

elzadj (approved):

moroz-art-dev (approved):

myroslavve (approved):
Migrator works nice, just one thing

Fybex (approved):

PR #3849, opened 2022-10-30
Issue 3809 xxl default values

elzadj:
# Description

Looks like a too easy fix  :thinking: 

Fixes #3809 

# How Has This Been Tested?

Need to test default attributes for AxisInput (padding/ margin / position), especially when the base is XXL. 

# Test checklist

**_ Front/Back Testing _**

-   [ ] Same 1-3 points on responsive

**_ Pre-Code Testing _**

-   [ ] Same 1-3 points on responsive
-   [ ] Test with 2 blocks of the same type-   [ ] Check no commented code and no unnecessary imports

# Checklist

### DIFF
--- a/src/components/axis-control/index.js
+++ b/src/components/axis-control/index.js
@@ -55,7 +55,6 @@ const AxisInput = props => {
 		label,
 		target,
 		singleTarget = null,
-		getValue,
 		getLastBreakpointValue,
 		breakpoint,
 		disableAuto,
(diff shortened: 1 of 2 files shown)


Naaaaiix (commented):
I'd say this situation is practically unique, but may appear again in future. Basically it comes from:
https://github.com/maxi-blocks/maxi-blocks/blob/2ba9a5f7fc7b9b313c1aa93989eacc72188eec1b/src/blocks/button-maxi/attributes.js#L146-L161

where we've got different `XXL` and `general` default attributes. In this case, the attributes are used by AxisControl component, but I think we should fix the attributes or the attributes mechanism, not the component. I say this because, even considering this implementation fixes this concrete situation, is not creating a fix for future situation where the attributes will be used by other components. 
Some other ideas that come to my head that may help fixing this one are:
1. On that attributes, instead of setting `general` set `xl`.
2. Change `getLastBreakpointAttribute` method for cases like this one 🤔

PR #3887, opened 2022-11-03
Issue 3755 - Default buttons incoherence in IB

Fybex:
# Description

Major changes:
- `border-style-general` - `none` instead of `undefined` by default now
- added the `isIB` argument, so now to create a special condition in the style helper for IB it is enough to destruct `isIB` and write the condition (for example, as it was done in `getBorderStyles` and `getBoxShadowStyles`).

Fixes #3755 

# How Has This Been Tested?
Tested if border and box-shadow default buttons in IB with some set attributes in block to affect is possible to turn off, tested if after changes, border and box-shadow settings is working as usual

# Test checklist

**_ Front/Back Testing _**

-   [ ] Test border default buttons in IB (check video from issue)
-   [ ] Text box shadow default buttons in IB
-   [ ] Test border default buttons in border settings
-   [ ] Test box shadow settings if styles applying as usual

**_ Pre-Code Testing _**

-   [ ] Test border default buttons in IB (check video from issue)
-   [ ] Text box shadow default buttons in IB
-   [ ] Test border default buttons in border settings
-   [ ] Test box shadow settings if styles applying as usual


Olekrut (approved):

elzadj (approved):

Naaaaiix (commented):
The implementation works good, but I'm not completely sure if this is the best way to do it. It may consume many processes and looks is re-rendering from outside the components with not a major necessity. Please @Fybex, let's have a look together and see if we can find a better fix 😸 👍

MeltingWings1 (approved):

Naaaaiix (changes requested):
Been deeply thinking on this implementation after yesterday's call with @Fybex to clear some doubts I had about how we are dealing with this situation. I'd like to share my point and request other devs what they think, cause part of the way we are resolving this situation is going to mark some norms/standards we'll need to follow in future situations.

Basically what this implementation is partly adding is an automatic transition effects for border when hover or IB is affecting. Let's have next situation in mind (it can be applied to hover or IB, going to explain it based on hovering the block):
We've got a block with a `border-style = none` and then enable border hover, and add a `border-style = solid` with a `border-width = 2px`. The result we had on `master` that responds to developing logics is the next one:

https://user-images.githubusercontent.com/50477222/204769300-e80b5ec1-37a9-4789-ab2f-865ec3a921dc.mp4

The border has a transition when hovering in, and the border-width gets bigger smoothly; but when hovering out the border directly disappears without animation or transition. This is easy to understand: on hovering in the `border-style` is 'solid' and uses the transition, and when hovering out the `border-style` gets back to 'none', so there's no transition possibility. If we would like to have a transition we just need to go the normal border and add a `border-style` equal to 'solid' and a `border-width = 0` 👍 

What this implementation is doing is to add that `border-style = solid` and `border-width = 0` automatically, so having the result is that we are showing on the sidebar a `border-style = none` but we are rendering `border-style = solid; border-width = 0`. So, the result would look like:

https://user-images.githubusercontent.com/50477222/204771170-c9b012a8-bcd0-4789-b73d-1559cac5a767.mp4

So, some of the issues I'm seeing on this implementation related to this topic:
1. Adding an iterator as it is `goThroughMaxiBlocks`on the helper `getBorderStyles` duplicates the average time we need to get the styles. To test it, I've measured a pattern from devdemo on this branch and on `master`. The average for both branches have been '0.21679ms' and '0.11846', respectively. 
2. The behaviour explained before generates a precedent for other hover situations. Imagine a background layer where creator adds a background-color just when hovering: to have correct transitions' creator should add the layer on normal (not-hover), reduce, for example, the opacity to 0 and then on the hover layer set it to 100 to have a transition on the opacity, in this case. So, following what has been done here, we should add CSS to ensure that opacity has transition even when creators didn't request it; so this sets us in front of a huge situation: we need to check every possible hover setting to ensure we are giving a smooth transition. Not sure if that's something we want to deal with.
3. We are removing the option for creators to have effects like the showed on the `master` version. I mean: with the situation we've got on `master` creators have the "freedom" to have that blinking hover effect where the border disappears abruptly while hovering off, or can add the commented settings to make it smoothly. With what has been suggested on this implementation, just a second option will be there.

Tbh, if it would be just the last point as an opposite argument I wouldn't be writing this and would be expecting testers feedback, but what concerns to me are points 1 and 2. There are other things I'm not completely sure they are correct, like re-rendering a block from a different component using [these lines](https://github.com/maxi-blocks/maxi-blocks/blob/e2627eb9d7c5a28f8e37a9ff5d629d3520e14339/src/components/relation-control/index.js#L603), but even with that we could live. But I'd like to know your opinion @elzadj @Marko64 @TeksPeks @MeltingWings1:
Should we add this automatic smooth transitions or leave the creators all options? And in case we want automatic transitions, do we want them like this? Should we apply to all hover options? 🤔 

Thanks for your patience and time 🙏

Naaaaiix (approved):
Thanks @Fybex 😸 

Check the comment left and can be merged 👍

PR #3896, opened 2022-11-07
Issue 3809

elzadj:
# Description

Fixes #3809

Need to test default attributes for AxisInput (padding/ margin / position), especially when the base is XXL.

# Test checklist

**_ Front/Back Testing _**

**_ Pre-Code Testing _**

# Checklist


Marko64 (approved):
I'm a bit lost how breakpoints and attributes work, but these inputs do seem to work properly on all breakpoints.

moroz-art-dev (approved):

Olekrut (approved):

Naaaaiix (commented):
As commented above, this situation is like the one we've got on #3881: an exception. In that case, the situation was:
`{
    attr-xxl: 'px',
    attr-xl: 'px',
    attr-l: '%',
}`
So added `attr-general: undefined` to fix it. It seems is a way that creates a good flow for creators, intuitive at least. In this concrete case we've got in front we've got:
`{
    attr-xxl: 23,
    attr-general: 15,
}`

So, let's imagine next situation: while having baseBreakpoint on `XXL` and resetting from there, what should happen? There are some possible options:
1. `attr-xxl` returns to its default value and leave `attr-general` as it is.
2. `attr-general` gets `attr-xxl` default value (baseBreakpoint is 'XXL', so it should save on general) and `attr-xxl` becomes `undefined` (as it's the same value than `attr-general`).
3. `attr-general` returns to its default and `attr-xxl` also returns to its default.

In terms of logics, 2 is for me the one that fits better with the current mechanism, but not sure if it's the best in terms of user experience.
Anyway, what I'm finding in this implementation is that while resetting on XXL basebreakpoint, XXL attributes get General default values, and Default values receive XXL values. The result is a small button on XXL and a big button on General:

https://user-images.githubusercontent.com/50477222/201059235-54f90103-b188-4cb6-bd34-61b1989068c7.mp4

I'm sharing all this text cause I don't have an idea of what's the correct answer lol but I think the one of the video is not going to be a good experience for creators. What may work is to return the iterator of line 568 (part of `onReset` function`) to its previous state based on `getDefaultAttribute` as I think the [changes done](https://github.com/maxi-blocks/maxi-blocks/pull/3881/files#diff-e39d67233a64857bf54cf7156c189fa897bca321b980fc35c870ef87fa5899ad) on #3881 would generate a result like:


https://user-images.githubusercontent.com/50477222/201061874-87c2033a-cd57-43f3-b743-9bdb8f09222b.mp4

🤔

PR #3853, opened 2022-10-31
Issue 3841 - Gap between blocks

Marko64:
Adds a constant margin of 50px in all blocks that are first on hierarchy (the ones that aren't wrapped by any other block) to allow leaving some space between them and letting the blue button to appear and add new blocks in the middle. Also made Container's indicators just appear when are selected, and added a label on them to difference `margin` and `padding`

Fixes #3841

PS: tests will break for sure, added an inline style for all blocks so snapshots will be like 🎆 

# Test checklist

**_ Front/Back Testing _**

-   [x] Make sure if you can click the block inserter between the blocks.
- [x] Make sure you can click the block inserter after set some margin to the blocks

**_ Pre-Code Testing _**

-   [x] Make sure if you can click the block inserter between the blocks.
- [x] Make sure you can click the block inserter after set some margin to the blocks
-   [x] Check no commented code and no unnecessary imports
-   [x] Standards of the project have been followed
-   [x] No errors/warnings on console

# Checklist

-   [x] My code follows the style guidelines of this project
-   [x] I have performed a self-review of my code
-   [ ] I have commented my code, particularly in hard-to-understand areas
-   [x] My changes generate no new warnings/errors
-   [ ] I have added/updated tests that prove my fix is effective or that my feature works
-   [ ] New and existing unit tests pass locally with my changes

PR #3903, opened 2022-11-08
Issue 3899 - Scrollbar on canvas with full-width blocks

Naaaaiix:
# Description

Removes padding issues on Maxi blocks. To understand the situation we need some context: Gutenberg editor adds some padding on the canvas (8px on both X axis sides), and then blocks have 2 subsequent styles:
1. Margin-left and margin-right === -8px
2. Width === calc(100% + 16px)

We can celebrate such a good practices like those, but anyway we need to adapt to it. The fix we've implemented was to modify the padding of the canvas of the Gutenberg editor, which affected the editor itself. The problem has been on setting that padding to 0: a scrollbar appeared on the bottom, and some native Gutenberg blocks were affected by that change we did. So, basically, we needed a solution that just affects Maxi blocks.

Considering that as a difference with native WP blocks, Maxi blocks has dynamic styles that affect margin and width, we can't rely as they do on static CSS. Also, our style mechanisms doesn't have support for having some styles on back and some on front, so wasn't a good solution. So, finally, I decided to take the same procedure done on #3853 to implement it; basically, added that styles as element inline-styles from the `MaxiBlock` component and just affect backend. 

Fixes #3899 

# How Has This Been Tested?

Create different blocks and set them on full-size. Not just Maxi, also native WP blocks. No scrollbar should appear on the bottom (unless we select a block with resizer handlers, that is an exception!)

# Test checklist

**_ Front/Back Testing _**

-   [ ] Test different blocks, Maxi ones and native WP ones, on full-width
- [ ] Test different responsive stages
- [ ] Test different themes

**_ Pre-Code Testing _**

-   [ ] Test different blocks, Maxi ones and native WP ones, on full-width
- [ ] Test different responsive stages
- [ ] Test different themes

# Checklist

### DIFF
--- a/src/components/block-resizer/index.js
+++ b/src/components/block-resizer/index.js
@@ -95,7 +95,7 @@ const BlockResizer = memo(
 				handleClassName,
 				showHandlesClassName,
 				!isCorner ? sideHandleClassName : cornerHandleClassName,
-				`maxi-resizable__handle--${axis}`
+				`maxi-resizable__handle-${axis}`
 			);
 
 		const handleClasses = {
--- a/src/components/block-resizer/index.js
+++ b/src/components/block-resizer/index.js
@@ -119,7 +119,6 @@ const BlockResizer = memo(
 				enable={enable}
 				handleClasses={handleClasses}
 				handleWrapperClass={handlesWrapperClassName}
-				// handleStyles={handleStyles}
 				onResizeStop={(e, direction, refToElement, ...rest) => {
 					onResizeStop?.(e, direction, refToElement, ...rest);
 					if (cleanStyles) stylesCleaner(refToElement);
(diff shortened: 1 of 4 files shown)


Olekrut (approved):

Marko64 (approved):
Imported several fullwidth patterns and tested with 2020, 2021 and 2022 themes.

The horizontal scrollbar is visible only temporarily now. Not the end of the world if it stays like this, but ideally it should not show at all:
Screenshot_3

elzadj (approved):
Just a small proposition - maybe make 8 a constant (and calculate 2x for 16)? Then, if Gutenberg changes it, we can fix it faster just in one place.

moroz-art-dev (approved):

PR #3989, opened 2022-11-28
Issues 3809 && 3945 - Resetting when winBase === XXL && `undefined` attributes become valued after reloading

Naaaaiix:
# Description

This branch aims to fix these 2 issues:
1. The commented UX on resetting when the padding for Button Maxi (follow the discussion on this PR #3967).
2. The uncomfortable issue we were finding on #3945 that doesn't allow `null` as a valid and different value than `undefined`.

As both issues were having troubles with the resetting, found a way to fix them both at the same time. The main implementation is based on 3 points:
1. Removed different default values on `attributes.js` that are affecting other responsive attributes than `general` and moved into `blockData` with the label `maxiAttributes`.
2. Added a new system on `maxiBlockComponent` class that, in case that new attributes exist, they are implemented on the first creation of the block overwriting the WP default attribute.
3. Added a new handle for the resetting called `handleOnReset` and a new capacity to `getDefaultAttribute` to iterate that new `maxiAttributes` object.

This way, as for WP that attributes doesn't have a default value, they can stay `undefined` in case we save like that, so problem number 2 is solved. In the other side, the issue we were facing on resetting from different responsive stages of the Button Maxi padding is solved, as now XXL and XL values are decoupled and can be reset easily. It may not fix 100%, but enhanced for sure. Once seeing the necessity and the flow of the creators, we can take advantage of the new `handleOnReset` function to even go further and improve the experience for that concrete cases/exceptions. 

Let's see if this works out and we can move forward with these issues 👍 

PS: this branch is a mix of the one pointing to #3967 and the other presented and enhanced on [this comment](https://github.com/maxi-blocks/maxi-blocks/issues/3945#issuecomment-1325472961). So, if it's merged, the rest of PRs can be closed 👍 

Fixes #3809 && #3945

# How Has This Been Tested?

It must be tested in many different ways:
1. Test modifying and resetting from different breakpoints the `Width/height` options of Container and Row Maxi
2. Test modifying and resetting from different breakpoints the Padding of Button Maxi
3. Test modifying and resetting from different breakpoints SC
* to all of these points, save and reload and check the attributes are correctly saved

# Test checklist

**_ Front/Back Testing _**

-   [ ] Test the points indicated above

**_ Pre-Code Testing _**

-   [ ] Test the points indicated above

# Checklist


Olekrut (changes requested):
image

need migrator for buttons which had padding:0px;

elzadj (approved):

Olekrut (changes requested):
[reset for paddings are working not correct.webm](https://user-images.githubusercontent.com/88893350/205078057-387f5f3a-8f21-495e-89d6-59858fa9c2c1.webm)

[button paddings.txt](https://github.com/maxi-blocks/maxi-blocks/files/10133202/button.paddings.txt)

MeltingWings1 (changes requested):

MeltingWings1 (approved):

Olekrut (approved):

elzadj (approved):

PR #3997, opened 2022-11-29
Issue 3958 - Background position after reset

MeltingWings1:
# Description

Removed the breakpoint from getDefaultAttribute in axis-control. It gets undefined as the default value for unit which breaks the input.

Fixes #3958 

# How Has This Been Tested?

1. Add different maxi blocks (text-maxi, container-maxi, etc)
2. Add a background layer and select set seperate
3. Change one of the position fields and then reset
4. Change it again and it should still update on the screen

# Test checklist

**_ Front/Back Testing _**

**_ Pre-Code Testing _**

# Checklist

### DIFF
--- a/src/blocks/text-maxi/attributes.js
+++ b/src/blocks/text-maxi/attributes.js
@@ -143,6 +143,7 @@ const attributes = {
 	...attributesData.zIndex,
 	...attributesData.customCss,
 	...attributesData.flex,
+	...attributesData.backgroundColor,
 };
 
 export default attributes;
--- a/src/blocks/text-maxi/attributes.js
+++ b/src/blocks/text-maxi/attributes.js
@@ -143,7 +143,6 @@ const attributes = {
 	...attributesData.zIndex,
 	...attributesData.customCss,
 	...attributesData.flex,
-	...attributesData.backgroundColor,
 };
 
 export default attributes;
--- a/src/components/axis-control/index.js
+++ b/src/components/axis-control/index.js
@@ -574,12 +574,7 @@ const AxisControl = props => {
 					customBreakpoint ?? breakpoint
 				)
 			] = getDefaultAttribute(
-				getAttributeKey(
-					getKey(key),
-					isHover,
-					false,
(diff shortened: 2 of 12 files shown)


Naaaaiix (commented):
The idea fixes the issue, but breaks other parts:

https://user-images.githubusercontent.com/50477222/204488759-3b68c2b5-c9cb-44a6-8a79-16ec088b0d5d.mp4

The change on AxisControl affects other instances of the component. Let me try to give some clues about this issue. The most important thing to understand with BackgrouldLayersControl is its attributes: instead of Gutenberg normal attributes, what we've got is an ['array'](https://github.com/maxi-blocks/maxi-blocks/blob/5701aa355de196bc26b5c20d06722517cc8c2ba5/src/extensions/styles/defaults/background.js#L13-L18). The reason why is an array is that we can have infinite layers, and each layer will be an object inside this array. In that sense, WP native function to get default block attributes, `getBlockAttributes` ('@wordpress/blocks'), can't get the defaults of each layer as they aren't registered into the block `attributes.js` file, so the source is other. In this case, the source of each layer comes from `src/components/background-control/layers.js`. So, resuming:
1. Removing the breakpoint on the `onReset` function of AxisControl breaks other instances of the component.
2. The source to retrieve the default attribute is different from the usual one, so we need to find a way to let BackgroundLayersControl to reset from it's default layers values 👍

Marko64 (approved):

Olekrut (approved):

elzadj (approved):

Naaaaiix (changes requested):
The issue has been fixed, but I think this implementation has an error: is delegating the responsibility of BackgroundControl reset into AxisControl. To keep a clean code strategy, each component should try to be as isolated as possible. Please, let's find another solution for it; one that can work can be to add a `onReset` prop on AxisControl that accepts a function, and in case it exists it is run instead of the original method used on AxisControl. That way the responsibility gets back to BackgroundControl and AxisControl keeps being an isolated and independent component 👍

PR #4016, opened 2022-12-01
Issue 4000 - SC typography

elzadj:
# Description

Fixes font-size, line-height, etc. in the SC Editor.

Fixes #4000

# How Has This Been Tested?

Checked SC values change for button, paragraph, headings.

# Test checklist

**_ Front/Back Testing _**

-   [x] Test the SC Typography for different blocks 

**_ Pre-Code Testing _**

-   [ ] Test the SC Typography for different blocks 

# Checklist


Marko64 (approved):

MeltingWings1 (approved):

myroslavve (approved):

Naaaaiix (approved):
It fixes the issue, but has a side effect we need to consider: it makes the difference with block attributes. I mean: when in a block attribute the value for XL, L and M is the same, we automatically remove values for L and M and just keep XL; on SC, with this implementation, we are going to keep them all. It means it will be a gap and incoherence between the way we deal with the SC attributes and the way we deal with the block attributes, and that can confuse creators. And not saying is wrong, just we need to be sure is what we need, and we accept it 👍 

For the rest, in terms of code, it has my approval! 🥳

PR #4023, opened 2022-12-02
Add IB clipPath support for Image Maxi

Naaaaiix:
# Description

After talking with @Olekrut about another issue he found, we played together with [this pattern](https://devdemo.maxiblocks.com/pure-image-light-pil-pro-part-2/) (number 12):

https://user-images.githubusercontent.com/50477222/205301625-ed374a6e-ecff-4790-bafb-dee84b65b994.mp4

To make this awesome effect on the images, the pattern is using a complex system. Here's the list view:
```
1. Column Maxi
    1.1. Group Maxi
        1.1.1. Group Maxi (IB trigger for the Divider Maxi effect)
            1.1.1.1. Image Maxi
            1.1.1.2. Divider Maxi (absolute positioning; IB makes it to get the border that create the effect)
        1.1.2. Group Maxi
            1.1.2.1. ... (Title, line, text, etc..)
```

The result couldn't be done in other way with tools we've got, so is awesome a pattern like that has been done: is a sign of huge knowledge and incredible creativity, but it still have some issues creators will struggle:
1. As Divider is an absolute positioned element, it's hard to access to the image as it overlaps it.
2. It's strange that a Divider block is set there without divider element, just like an empty layer. That's really confusing.
3. Divider was using customCSS for the width and the height.
4. We are creating, IMO, many unnecessary elements.

So, considering that I wanted to try a solution. My goals were:
1. Remove Divider but keep the effect.
2. Reduce the amount of elements preserving IB effect and styles.
3. Make the pattern easier for creators.

After trying hard I didn't find any other way than the proposed there, but I saw the light with a new support on IB: ClipPath. Here's a video with the result, where the first column shows the old way, the middle column is an experiment that doesn't work, but the third one is doing exactly the same as 1:

https://user-images.githubusercontent.com/50477222/205303814-423e62e1-4220-4fa3-86b7-a2755b115b3b.mp4

With this fix, the structure of the Column now is:
```
1. Column Maxi (has the IB trigger responsibility)
    1.1. Image Maxi (uses clip-path `inset`; 0 on normal state, 20 on IB)
    1.2. Group Maxi
            1.1.2.1. ... (Title, line, text, etc..)
```

So, we removed 4 blocks; that's great, easier and simpler for creators, lighter for downloading, fewer elements on DOM... awesome! And also, we improved:
1. User experience: now is easier to click on the blocks and select them.
2. User experience: there isn't an empty Divider block any more.

# How to test it?

Here's the [code editor](https://github.com/maxi-blocks/maxi-blocks/files/10141458/clipatib.txt); first column has the old system, third column the new one. Both should look and do the same effects 👍 
Also, test `clip-path` from IB targeting an Image Maxi 👍 

# Test checklist

**_ Front/Back Testing _**

-   [x] Do the tests commented

**_ Pre-Code Testing _**

-   [ ] Do the tests commented

# Checklist


Olekrut (changes requested):
1) when we're enabling clip-path in IB ( don't do it first for image-block ) transition is not working - i guess should be enabled for image-block - auto. Adding random values ( 5 and 15)

[clip-path.webm](https://user-images.githubusercontent.com/88893350/205906787-5eb24f2b-67db-4cd7-b370-99da72686c57.webm)

2) Need units for clip-path values, not only %, need PX for sure, and maybe EM

Marko64 (approved):

Olekrut (approved):
Will wait fixes in another issue

elzadj (approved):

MeltingWings1 (approved):

PR #4042, opened 2022-12-07
Text Maxi when pressing enter

Naaaaiix:
## Description

This implementation was talked on a call with @ckp267, and has been growing out of Github and issues; so I'll explain it here:

The main implementation of this branch is changing the way we deal with Text Maxi when pressing enter. The old way you can find on `master` makes a break-line when pressing enter inside of a Text Maxi, but we've found some UX problems:
1. On pasting text, everything gets into the same block, which makes it hard to then get some titles and transform them into Headers.
2. It was hard to keep the break lines.
3. It was completely opposite to the default Gutenberg's experience, which can be confusing for our creators.

With that, we've been testing different options and here are the implementations done in this branch:
1. Now, on pressing enter inside a Text Maxi, a new Text Maxi is created.
2. On pasting, each paragraph or title is set in different blocks.
3. To deal with the margin we need between Text Maxi blocks, and to don't have conflict with the margin of the block, we created a new attribute/setting called `gap-bottom` that affects the inner element of the block (the paragraph, heading, list...). This feature affects Text Maxi, Image Maxi caption and hover-effects.
4. Implements `white-space` requested on #4098 and `word-spacing` features on Typography and SC.
5. Set `Outside` as the default list-style on list mode of Text Maxi.


Olekrut (commented):
https://jam.dev/c/f189350b-7b30-480c-901a-dedc56cbd12e
 is it ok, that instead breaks are adding a few text-blocks?

Olekrut (approved):

elzadj (approved):

ckp267 (commented):
Good to go. Just spotted an issue with bullet points if text drops to second line. Josh said it should make a new issue for it.

PR #4083, opened 2022-12-14
Load pattern and Select icon buttons

Rustam198804:
# Description

Fixes #4031


elzadj (changes requested):
Need to fix that weird ```test``` class :+1:

Marko64 (changes requested):
1. I think icons need to be the same colour:
Screenshot_1


2. Perhaps we can create a variable for these repeating colours:
Screenshot_2

Olekrut (changes requested):
https://jam.dev/c/3785d077-a678-4609-a2a1-34c0e96dbbca
Check, how it looks for narrow columns

Olekrut (changes requested):
https://jam.dev/c/711a8bc6-065d-4806-891f-124d54e2206d
1) need hover for smile icon
2) why when we pick M we have Icon and text, but when we pick S and then M, we have only icon
3) when we pick M, icon is compressed, distorted
image

Olekrut (approved):
See only icon from L, so it is fixed

myroslavve (approved):

elzadj (approved):

PR #4116, opened 2022-12-22
Issue 3294 - Removed duplicate fit content from column

Marko64:
Removed the duplicate `fit content` option for columns.

Our only concern here was if the removed option was used in patterns. Checked with @Olekrut, and he said that we used the other option in most cases because the one we removed wasn't even working. 

Fixes #3294 

# Test checklist

**_ Front/Back Testing _**

-   [ ] Make sure that the fit content option under height/width is removed for column, but that it's still there for row and container.

**_ Pre-Code Testing _**

-   [ ] Make sure that the fit content option under height/width is removed for column, but that it's still there for row and container.

# Checklist

### DIFF
--- a/src/blocks/column-maxi/inspector.js
+++ b/src/blocks/column-maxi/inspector.js
@@ -79,6 +79,7 @@ const Inspector = props => {
 										block: true,
 										hideWidth: true,
 										hideMaxWidth: true,
+										hideFit: true,
 									}),
 									...inspectorTabs.marginPadding({
 										props,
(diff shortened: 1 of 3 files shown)


Naaaaiix (approved):

elzadj (approved):

Olekrut (approved):
Awesome) Just a question - Creators would not be confused that for all blocks we have Fit content in tab Height/width and only for column in tab Column size, for me is not a problem and it is also having logic

PR #4455, opened 2023-02-28
Optimization - Reduced attr labels prefix fix idea

Fybex:
# Description

Size settings don't work with attributes, which have prefixes in #4449 . It's because for example in `getAttributeValue` firstly we're calling `getAttributeKey` and then `parseLongAttrKey`, but on `parseLongAttrKey` after clearing for example attribute key: `button-width-general-hover` we will have `button-width`, so I suppose we will need:
1) to move `parseLongAttrKey` logic inside `getAttributeKey` and not include `prefix` in `parseLongAttrKey` call.
2) add to `getLastBreakpointAttribute` `prefix` param to define exactly where `target` and where `prefix`
3) don't modify object with`parseLongAttrKey` in `handleSetAttribute`, we should do it straight away in every `onChange`, because we won't be able to define `prefix` in `handleSetAttribute`

This PR tries to deal with this.

# How Has This Been Tested?
Checked if size settings works for prefix attributes, for example for button-maxi width(not canvas!)

# Test checklist

**_ Front/Back Testing _**

-   [ ] Test size control with prefix attributes(button-maxi width/height for example, not canvas!)

**_ Pre-Code Testing _**

-   [ ] Test size control with prefix attributes(button-maxi width/height for example, not canvas!)

# Checklist

### DIFF
--- a/src/components/full-size-control/index.js
+++ b/src/components/full-size-control/index.js
@@ -10,9 +10,10 @@ import AdvancedNumberControl from '../advanced-number-control';
 import ToggleSwitch from '../toggle-switch';
 import withRTC from '../../extensions/maxi-block/withRTC';
 import {
-	getLastBreakpointAttribute,
-	getDefaultAttribute,
+	getAttributeKey,
 	getAttributeValue,
+	getDefaultAttribute,
+	getLastBreakpointAttribute,
 } from '../../extensions/styles';
 
 /**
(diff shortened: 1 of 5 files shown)


Naaaaiix (commented):
The suggestion deals with the problem, but I don't think it does it in the correct way in the current circumstances. As I've already shared with @Fybex on chat, this suggestion reduces the focus on having few changes that affects the totality and puts more efforts on refactoring components. That could seem cleaner, but the amount of work and the risks of breaking things increases a lot. In a refactoring like we are doing, we need to do baby-steps and ensure we do them with security. 

Pushed a fix for the issue commented on this PR on [this commitment](https://github.com/maxi-blocks/maxi-blocks/pull/4449/commits/ae5d723146c64e44e159382624fde7b9cada827c). Closing the PR, but keep the conversation if consider necessary 👍

PR #4430, opened 2023-02-24
Issue 3388 - DC

Naaaaiix:
# Update 3/3/23
This PR is not for Text Maxi any more, now it includes DC for:
- Text Maxi
- Button Maxi
- Image Maxi
- Link component

🥳 

# Description

This PR replaces #4215

Basically does the same as the previous PR: it gives DC support to Text, Button and Image Maxi, and Link component. Changes done in this branch:
1. Replaces 'apiFetch' calls with 'entity' calls (WP method for Gutenberg)
2. Removes DC JS file from frontend
3. Adds e2e simple tests
4. Simplified the supported fields
5. Adds a new HOC for all future DC blocks (`withMaxiDC`)
6. Add support to update the content when loading the EE
7. Adds DC support as 'inspectorTab' to appear in settings sidebar
8. Creates DC folder with function and DC logics for future supported blocks
9. Reduced the re-renders
10. Increased the speed
11. Fixes errors

These are the expected actions you can do with DC depending on the block:
- [Text Maxi](https://docs.google.com/drawings/d/1G3LXxwKBRRQJUhSv81hcj3h3nHKlDbFuMqkW3WjnxNU/edit?usp=share_link)
- [Button Maxi](https://docs.google.com/drawings/d/1PRdr6TNC9JSB63eR7T3S_nqX-vwQ7Fx9L1B-WasI3K4/edit?usp=share_link)
- [Image Maxi](https://docs.google.com/drawings/d/13CFNIc1PLrQK-RukvJrlGIjvohNGuQ-IrqENLxB29UI/edit?usp=share_link)

Other actions and field support will be added in future, these are the simplest and most important ones. If a new one is necessary, please think on sharing a diagram like the one above.

Some known issues:
1. Styles might need some love
2. Using custom-date, the frontend is not exact sometimes. We need to reduce options or fix it somehow 👍 
3. On changing the type it returns a console error sometimes; doesn't affect the functionality
4. On Image Maxi, alt and caption components on the sidebar need to be checked. Didn't fix it as it need deeper look and I prefered to don't mix topics.
5. When enabling the link from DC on Link component of the Toolbar, you'll need to click out and it will update. It's an issue on the component that needs to be fixed apart.
6. Not sure how to deal when blocks like Image doesn't have any content on frontend, will fix it next week.
7. Part of the code needs to be cleaned, like 'dc-error' 👍 
8. Need to check the relation of Text Maxi with List options when DC.
9. It's probable that some options get in conflict with DC. Share the ones you find please.

[...]

PR #4469, opened 2023-03-02
DC for Button Maxi

Naaaaiix:
# Description

This branch enhances #4430 and extends DC universe (lol) including support for Button Maxi. It has the same errors than the other PR cause it's an extension of it, so please, try to don't duplicate problems here and there. Here you can find the diagram of the support options; remember we'll be adding new ones as soon as DC will be ended:

[DC - Button Maxi.pdf](https://github.com/maxi-blocks/maxi-blocks/files/10869398/DC.-.Button.Maxi.pdf)

Also fixes a minor issue: the number of possible elements to choose was limited to 10, now should get them all.

No tests added, will implement later 👍 

# How to test it?

Test DC for Button Maxi 👍


Olekrut (changes requested):
1) not sure that related with this brunch - popover is overlapping button
image
2) long names
image

PR #4520, opened 2023-03-09
Blocks preview

kyrapieterse:
Bye


Olekrut (changes requested):
removed

Olekrut (approved):

Olekrut (changes requested):
Bye

Olekrut (changes requested):
Slide preview and error after https://jam.dev/c/98041a44-5c8d-4af1-8b06-aa6ab070e6c3

darimshlk (changes requested):
removed

darimshlk (approved):

Naaaaiix (commented):
Cool, works fine; I'd like to do a suggestion: instead of adding the preview situation to each block, let's unify everything into a HOC. It's easier to implement and reduce the amount of code, and also reduces the load and makes the changes on the preview faster. Made a implementation into Group Maxi in [this branch](https://github.com/maxi-blocks/maxi-blocks/tree/block-preview-2), just need to be implemented on the rest of blocks 👍

Olekrut (approved):

PR #4547, opened 2023-03-14
Optimization - accordion icon

myroslavve:
# Description
Accordion icon group adds no new attributes, so I just removed it to not create confusion.

# How Has This Been Tested?
Checked that icon settings work fine.
Icon size doesn't work for the reasons I shared in chat (because width attributes are exported not shortened)

# Test checklist

**_ Front/Back Testing _**

-   [ ] Check accordion icon.

**_ Pre-Code Testing _**

-   [ ] Check accordion icon.

# Checklist


Naaaaiix (commented):
Apart of the possible confusion, is there any reason to move the `accordionIcon` out of default attributes folder? Not saying is wrong, just trying to see what can be more confusing: to have a group of attributes outside of the folder or to have the file without the reducer attributes function. My thought is: that function will end disappearing in phase 3, so the possible confusion will disappear. Maybe better to leave the file as it was and leave a comment saying why is not using the `attributesShorter` 👍

PR #4557, opened 2023-03-16
New block Loader

Naaaaiix:
# Description

This branch deprecates the `lazy` and `suspense` implementation and replaces it with a block loader. What do we expect of this implementation:
1. It's not an optimization, so things aren't faster. Its aim is to create the illusion that Maxi is faster.
2. Improves the UX (user experience): instead of spending time in front of an empty blank page that can even return some waiting advice (`the page doesn't respond, close or wait`), it makes Maxi blocks wait for the editor to be loaded, and then it loads the blocks. 
3. It doesn't make the load longer or slower: after many comparisons, we saw the average time between this branch and a branch without this implementation is the same. To test it by your own, you can use [this branch](https://github.com/maxi-blocks/maxi-blocks/tree/revert-suspense) (called `revert-suspense`). But the main point is: we offer some movements to creators that see changes on the screen while everything is loading, which at the end is much better than an empty page.
4. It just works when we load the page and there are saved Maxi blocks. So, in new pages or when downloading from the Library, this loader is disabled. Just works when loading a page that already has blocks; once loaded the content, the loader is disabled.

# How Has This Been Tested?

The first test is to compare metrics. Here's a video between this branch and `revert-suspense` branch: 

https://user-images.githubusercontent.com/50477222/225551497-9cecd2fc-1400-4205-8302-a46202c15ae0.mp4

(This video was recorded for @kyrapieterse and @ckp267, so it's out of context)

Then, other things checked:
1. The loader just works when the page already contains Maxi blocks
2. On a new page, the loader is not appearing when adding blocks
3. When adding from the Library or pasting code, the loader doesn't appear

# Test checklist

**_ Front/Back Testing _**

-   [ ] Test the aspects commented above

**_ Pre-Code Testing _**

-   [ ] Test the aspects commented above

# Checklist


Marko64 (commented):
I like this better. But one thing can get on one's nerves:
https://user-images.githubusercontent.com/19425503/225557739-b2b72e51-7089-4439-ad3e-dfd7625aacf1.mp4

Olekrut (approved):

darimshlk (approved):

elzadj (approved):
Bye bye suspense

Fybex (changes requested):

elzadj (approved):

Fybex (approved):

PR #4683, opened 2023-04-04
Optimization - IB and `goThroughMaxiBlocks`

Naaaaiix:
# Description

Tried to reduce the impact IB can have on Maxi speed, basically creating couple of stores that replace most cases where `getTrhoughMaxiBlocks` was running. This function was getting all blocks and checking one by one; with the a store for `relations` and a store for `blocks` that related `uniqueID`s and `clientId`s we can reduce the iterations and therefore the resources we are spending there. 
The optimization might not be hugely noticeable on the editor and its actions, and not going to make faster the opening of the view list or the blocks list, but you can be experienced by swapping to code editor. Basically it makes the unmount of the blocks faster (or that is supposed).

# How Has This Been Tested?

Apart of the speed measurements we can do, the parts modified has been tested too:
1. On adding an IB item, if the block changes the selected setting hover state, IB receives the change.
2. On removing a block that is target of IB, the IB interaction disappears.
3. When couple of blocks share same uniqueID, one of them is changed 👍 

# Test checklist

**_ Front/Back Testing _**

-   [ ] Test the points commented above

**_ Pre-Code Testing _**

-   [ ] Test the points commented above

# Checklist


Olekrut (changes requested):
https://jam.dev/c/98cb0e9c-6630-40d7-975b-c2d88d90718e

if to create IB, after that duplicate some of existed block>
1) will be affected new duplicated block
2) will be visible only new duplicated block

darimshlk (approved):

Olekrut (changes requested):
https://jam.dev/c/7bdf07d6-be3f-4acd-8321-158c1a5a6d0d
created blocks are adding to the end IB`s list 
only if re-upoad edit page list will have correct order

Fybex (changes requested):

Olekrut (approved):

PR #4698, opened 2023-04-07
Optimization - Remove main `subscribe`, remove `marginValueCalculator` observer and introduce `CSSCache` concept

Naaaaiix:
# Description

This branch includes several optimizations with small changes in order to avoid possible breaks. Anyway, it needs a deep testing as what has been changed are some functionalities of the core of Maxi:
1. Removed the main `subscribe` of Maxi from `dom.js`: these lines of code were really old and were using way many resources, much more than necessary. To remove it, it's part of code have been split and replaced in other locations basically under the `useEffect` hook.
2. Introduced the `CSSCache` concept, that saves in `maxiBlocks/styles' store all the responsive styles in case we swap from one responsive stage to other we can pick the styles from there and no need of re-render. This makes the opening of List View or Inserter or the responsive stage swap much faster.
3. Removed the `marginValueCalculator` particular 'ResizerOberserver' that was creating an observer for each block and replaced on `dom.js` to make it unique and centralized the value into a 'maxiBlocks/style' store, that way `maxiBlock` component gets the content from there and we reduce the amount of unnecessary observer instances.

# How Has This Been Tested?

The things we need to test are even more the ones I'm going to write:
1. The first point affects to:
- Toolbar
- Sidebar
- Swapping responsive stages
- Dragging blocks
- Adapting Maxi blocks styles on Gutenberg native responsive (careful #4703)

2. The second point can affect:
- Swapping responsive stages
- Open List View or Inserter

3. The third point can affect:
- On changing to different themes, the blocks should be always centered, no matter if they are fullwidth or normal width.

Everything needs to be tested on normal editor and FSE 👍


Marko64 (changes requested):
Switching responsive views using Gutenberg native options doesn't work correctly yet. Alignment and row width looks broken. Let me know if you need a video.
EDIT: We already have an issue for that.

Marko64 (approved):
Tested all points and didn't find anything not working but changing responsive using native Gutenberg options.

Dragging used to be slow, but now feels significantly faster.

Olekrut (approved):
Mach better) Thank you) will be need to implement similar speed for swapping between code and visual editor 🥇

darimshlk (approved):

elzadj (approved):
:dancers:

PR #4709, opened 2023-04-10
Issue 4695 - button padding on XXL

myroslavve:
# Description
Changed logic of `cleanAttributes` in particular `flatNewAttributes` to save new XXL attributes that are same as general, when that attribute is already defined.

Fixes #4695 

# How Has This Been Tested?
Checked that button padding can be changed on XXL both as base breakpoint and responsive breakpoint.

# Test checklist

**_ Front/Back Testing _**

-   [ ] Check that button padding can be changed on XXL as a base breakpoint.
-   [ ] Check that button padding can be changed on XXL as responsive breakpoint

**_ Pre-Code Testing _**

-   [ ] Check that button padding can be changed on XXL as a base breakpoint.
-   [ ] Check that button padding can be changed on XXL as responsive breakpoint

# Checklist


Marko64 (approved):

darimshlk (approved):

Naaaaiix (approved):

Olekrut (changes requested):
https://jam.dev/c/629e2685-2bd1-4062-9658-839e3b0149b2
I understand that video is long, but take a look

try to set padding for button ( for XL and XXL )  when base is XL > after that resave page with this button when base is XXL and check frontend on XL

Olekrut (approved):
I understand that it is issue only about button, but need to do the same for all blocks and settings 👍

Naaaaiix (commented):
Sorry for the huge delay on this PR, but wanted to check it deeper and haven't got the time. Let's see if we can get the best result together, cause we all know how complex is the attributes system. Already contacted @Fybex to help me discussing about it cause we need more than one pair of eyes for review 😸 

The first to say is that this implementation fixes the issue, but as always it can have some side-effects I'd like to check; doesn't mean it's wrong, just want to be sure we are all agree with what it adds. 
The old flow was, being in a baseBreakpoint of XXL and having these attributes
```
{
    test-general: 15,
    test-xxl: 50,
}
```
that when changing `test` attribute to 70, we were receiving
```
{
    test-general: 70,
    test-xxl: 50,
}
```
So, when jumping to XL responsive stage it shows `general` (70) and then when coming back to XXL it shows XXL (50), so seems is wrong (this can be checked on the video of the issue). The proposal of this implementation is, when changing `test` attribute to 70, we receive
```
{
    test-general: 70,
    test-xxl: 70,
}
```
The first issue I see is the duality: is there any reason to save same value for both attributes? Is not redundant and wasting code editor space? 🤔 
The second thing that comes to my head is the relation with lower attributes than XXL, from XL down: they were having another value before, that can be a bit strange. As @Fybex commented when we were talking about it, maybe in this concrete case what we need to have returned is:
```
{
    test-general: 70,
    test-xl: 15
}
```
and in case users want to change, they can do it from there. But this breaks a bit with the idea of do all changes on general and then do the fixes on responsives: we are modifying a responsive value from baseBreakpoint, that doesn't seem coherent to me, tbh.

So, maybe what I consider it can be the best solution is to return:
```
{
     test-general: 70,
     test-xxl: undefined,
}
```
That way, on modifying from XXL being baseBreakpoint (so `general`), we keep just the general value; once creators end with the general styles they get into smaller responsives and make them the fixes it needs.

Not sure about my suggestion, not sure about the implementation, and not sure what I'm going to dinner today. What I'm 100% sure is that we are going to do it better if we do together 😸 

As a side note, I'd like to share that this situation is just happening on Button Maxi padding options, and maybe can be related or used with `maxiProps` as it's related with it (I mean, we might need a conditional that checks `maxiProps` or does an exception in that cases, not sure).

I expect your ideas @TeksPeks and @Fybex, let's see if we find the solution 💪

PR #4708, opened 2023-04-10
Issue 4434: Fixed IB issue with Untouched settings

FitseTLT:
# Description
For IB in most cases, untouched settings were included into the IB in the frontend and there were Unwanted Differences between Editor and frontend

Fixes #4434 

# How Has This Been Tested?
Create IB and change some settings and try to go to 'Block to affect' and change the untouched setting in the IB and check that it follows the default setting of the block in IB (editor) and also check that it is reflected in the frontend

# Test checklist

**_ Front/Back Testing _**

-   [ ] Test the component in normal state in sidebar
-   [ ] Test the component in hover state in sidebar (if hover exists)
-   [ ] Check that the settings work on frontend
-   [ ] Check that backend works and saves the settings after the editor reload
-   [ ] Same 1-4 points for the toolbar
-   [ ] Same 1-4 points on responsive
-   [ ] Test the block inside the grid (Container + Row + Column)
-   [ ] Test the block as a standalone block
-   [ ] Duplicate the block, test that the settings of the first do not affect the second
-   [ ] Test with 2 blocks of the same type

**_ Pre-Code Testing _**

-   [ ] Test the component in normal state in sidebar
-   [ ] Test the component in hover state in sidebar (if hover exists)
-   [ ] Check that the settings work on frontend
-   [ ] Check that backend works and saves the settings after the editor reload
-   [ ] Same 1-4 points for the toolbar
-   [ ] Same 1-4 points on responsive
-   [ ] Test the block inside the grid (Container + Row + Column)
-   [ ] Test the block as a standalone block
-   [ ] Duplicate the block, test that the settings of the first do not affect the second
-   [ ] Test with 2 blocks of the same type
-   [ ] Check no commented code and no unnecessary imports
-   [ ] Standards of the project have been followed
-   [ ] No errors/warnings on console

# Checklist

-   [ ] My code follows the style guidelines of this project
-   [ ] I have performed a self-review of my code
-   [ ] I have commented my code, particularly in hard-to-understand areas
-   [ ] My changes generate no new warnings/errors
-   [ ] I have added/updated tests that prove my fix is effective or that my feature works
-   [ ] New and existing unit tests pass locally with my changes

Issue #3811, opened 2022-10-20
Implement new attributes saving mechanism to object attributes


Naaaaiix:
The new attributes saving mechanism helps having clean attributes and improves the creators experience. But it works correctly with all type of attributes that aren't object/array type. That ones need to be checked different, as it has been done with `background-layers`, where `handleSetAttributes` ensures the new mechanism working inside an attribute that is an object. We should implement this mechanisms for all attributes with object/array type, like they can be `relations`(IB), `transform` or `transitions` 👍

PR #4715, opened 2023-04-13
Issue 4434 - Implement `handleSetAttributes` to IB

Naaaaiix:
## Update 18/4/23
Some updates:
1. Fixed some issues related with the saving methodology with IB
2. Created unit tests

Some notes:
1. Need to clean the IB attributes when they are repeated; something similar to `styleProcessor`/`styleCleaner`

## Update 14/4/23
Some new updates and features:
1. Fixed the issue @Olekrut shared with responsive values
4. Created a system that updates IB styles from the affected block, which ensures the IB settings are always synchronized with the block to affect attributes.
5. Created kind of migrator on `maxiBlockComponent` that updates the old IB attributes.

# Description

This PR is a parallel and alternative version of #4708. As already commented to @FitseTLT, the #4708 PR is a correct and a really good implementation that fixes the issue, but in this concrete case uses a methodology that could be simpler having the correct context of the situation: #4434 issue is related with #3811, as the main problem is IB is saving sometimes more attributes than the ones it needs. If we watch the video @Olekrut shared on #4434, what we see is that we change a property not related with colour, then we change on the IB affected block the colour, but in IB, when triggering the action, the colour comes back to the old one. The reason is that we aren't saving just the modified attributes on IB in the `relation` object, we are saving all the attributes! So, following the requirements of #3811, this PR implements `handleSetAttributes` on IB to ensure we save the correct attributes. The list of implemented actions is:
1. Implements new filters on RelationControl that first cleans the attributes to save on `relation` object using `handleSetAttributes`, and to ensure a good result, ensures we save the related `unit` or `palette` attributes in case is necessary, and then gets the styles using a temporary object using the attributes defined on IB settings objects under the key `styleAttrs`.
6. As commented above, implemented a new entry on Relation settings object called `styleAttrs` that ensures that attributes will be used on the getting styles process for the IB cause normally this attributes are necessary inside the concrete helpers. These attributes aren't saved inside the IB object 👍 
7. Made a small esthetic change on the `src/extensions/styles/helpers/getAlignmentFlexStyles.js`, nothing big lol

PS: this implementation also works as an optimization, as it reduces the amount of attributes saved on `relations` object reducing the code-editor 👍 

Fixes #4434 and partially #3811 

# How Has This Been Tested?

To test this PR needs a bit of patience and love. The process is the same commented on #4434:
1. Create a Group Maxi and set inside a block to affect.
(*). Sometimes is necessary to set some attribute on the block to affect, like `box-shadow` options or button icon options
2. On Group Maxi set an interaction that contains colour, but don't change the colour, change other setting
3. Check frontend all work good.
8. On editor modify the main colour of the affected block (not from IB, from the affected block)
9. Check on frontend the colour is the selected one and the IB trigger doesn't change it
10. On Grup Maxi IB, change the colour of the block to affect
11. In frontend, with the IB action not triggered, the block should show the colour selected from the affected block, and on triggering the IB the colour should be modified to the one selected on the IB of the Group Maxi.

Here's a video:

https://user-images.githubusercontent.com/50477222/231695363-a83ee7c2-1562-43f9-99b1-1aed16a51f4c.mp4

# Test checklist

**_ Front/Back Testing _**

-   [ ] Test IB in general lol

**_ Pre-Code Testing _**

-   [ ] Test IB in general lol


FitseTLT (commented):
From testing perspective I think you have to check your implementation I found this:
1. for divider maxi: tried to change only color of box shadow and in the frontend all the other values were 0px and the shadow disappears in frontend(for IB).
2. also tested for it's canvas bg color layer tried to change its color but still didn't work.

To comment on the implementation I liked the idea of styleAttrs for attrs like line-orientation bg-active-media ... But I don't think it applies to all. There are case we need to save the attributes like in box-shadow. Because if we don't save the values in relations attribute the editor side will reflect the recent updated value of the block to affect but the frontend's IB is calcultated according to the old values which existed at point of the creation of the relation.

Things left: 
1. box-shadow attributes map should be added to other areas for eg. button box-shadow
2. background gradient opacity is also needed by the style helper should be added to style attrs.

Naaaaiix (commented):
Awesome, good job @FitseTLT 😍 
Left some minor aspects, but it looks really good. Now we need a deep testing of all IB settings to ensure we haven't break anything 😅 Once done, we might need to create a migrator 👍

Olekrut (changes requested):
https://jam.dev/c/14495b07-dd4e-470f-8380-eb07129add7c
unit is not changing on XXL after some manipulations on responsive
[unit.txt](https://github.com/maxi-blocks/maxi-blocks/files/11232483/unit.txt)

Fybex (changes requested):
Greaaat job overall @Naaaaiix & @FitseTLT 🎉 . Only a few comments.

Olekrut (changes requested):
1) Icon colour is not working for IB
image

the same with button`s icon 
image

2) https://jam.dev/c/c8fdc054-c908-4922-96fa-695c8d8cbbb5
clip-path is not working in IB for Image-block ( need to check for BG Image )

3) https://jam.dev/c/65d7e7d9-1915-438f-b43b-8019eeaf7d31
colour, values for box-shadow are not inheriting to IB after changes in block ( all blocks )

Olekrut (approved):

Olekrut (changes requested):
https://jam.dev/c/88c4e958-337e-43db-98a7-e316a66faf05

Olekrut (changes requested):
[new new new IB .txt](https://github.com/maxi-blocks/maxi-blocks/files/11250388/new.new.new.IB.txt)
units for XL are not changing if create IB when XXL is base

Fybex (changes requested):

Olekrut (changes requested):
are not working IB related with BG layers on existed patterns
for example, try download PTML-PRO-99 and 100

Olekrut (approved):

elzadj (approved):
wow, great (and massive) job

PR #4794, opened 2023-04-28
Issue 4642 - ACF DC integration

myroslavve:
# Description

Added ACF integration to DC.

Fixes part of #4642 

# How Has This Been Tested?
Before testing, you need to install ACF plugin.

Checked that ACF fields are selectable from text maxi, image maxi and button maxi,
here's how to create an ACF field and select it:

https://user-images.githubusercontent.com/53875005/235209772-8080f6f9-8ece-40ce-8baf-cd712c14adc9.mp4

# Test checklist

**_Front/Back Testing_**

-   [ ] Test text maxi, button maxi and image maxi acf integration.
-   [ ] Test blocks behaviour after acf plugin is disabled.
-   [ ] Check that only suitable acf fields are available in blocks (for example, only image field should appear in image maxi)
-   [ ] Check frontend.

**_Pre-Code Testing_**

-   [ ] Test text maxi, button maxi and image maxi acf integration.
-   [ ] Test blocks behaviour after acf plugin is disabled.
-   [ ] Check that only suitable acf fields are available in blocks (for example, only image field should appear in image maxi)
-   [ ] Check frontend.

# Checklist


Olekrut (changes requested):
have checked 15 points
working only 2 of 15
image
some of them is not displaying in list
image

Olekrut (approved):

elzadj (approved):
Approved, but would be great if @Naaaaiix review it too. It's a big one.

Naaaaiix (commented):
Good implementation, let's see if we what I share can improve the code quality and prevent some future issues 😸 

I think the API implementation is not necessary. ACF has it's own API where we can ping, but also we've got 2 others easier ways to GET and POST ACF information: the first, the well-known `getEntityRecords` that can directly remove the implementation done in the Maxi API; the second, which I consider even better, is to use the `acf` native package that can be used on backend. Here's the [ACF API for JS](https://www.advancedcustomfields.com/resources/javascript-api/). At the same time, this API contents some methods we already implemented. 

My suggestion: replace as much of our code by native WP/ACF methods. The reason is easy: they have better tools for these technologies than us, as its their product. Don't reinvent the wheel and use things supported by native technologies will mean less issues in future, lighter implementations and easier methods to keep growing and scaling the job.

Let's have a look with this perspective and once decided if we go with it or not, I'll check the rest of the PR and the implementation. Thanks 😸

elzadj (approved):

PR #4798, opened 2023-05-02
Issue 4544 - Fix `React root` issue from React 18

Naaaaiix:
## Update (4/5/23)
After checking the reviews done, decided to rollback the point 2 commented on the previous update that was modifying the way we are serving the templates and templates parts styles for blocks. I'll share the issues is generating and the reasons why I wanted to change in other issue 👍 
Thanks for the reviews 🙏 😸 

~~PS: found some broken things; drafting it 👍~~ => Nothing new broken, same in `master`. Will create new issues 👍  

## Update (3/5/23)
After checking deeper, did some changes to ensure a better performance of Maxi in EE but specially in FSE. To do it, the current implementation adds:
1. A new store called `maxiBlocks/blocks` that was already implemented on #4683 and offers the capacity to save the blocks with their root style element, that way it helps fixing the `createRoot` error when swapping from code to visual editor and saves the re-rendering of the styles, that ends with a optimization of the speed when swapping from code to visual editor.
2. ~~Modified the way we serve the styles from template parts: the current implementation was using the `uniqueID` to set the CSS target specificity for the blocks in template parts. The new implementation avoid using the `uniqueID` as it was a problem when moving the blocks from a template part to another template part or a page/post, as the `uniqueID` kept with the original template name. The new implementation gets the `tagName` of the current template part and adds it in the CSS we are creating, that way we can preserve the `uniqueID` isolated of the context.~~
3. Added new triggers that returns the `isPageLoaded` to `false` as it is when we load the page for first time. This way we optimize the UX in situations like when we swap from code editor to visual editor, or when we pass from the FSE general editor to a template part. Modifying this global variable and returning it to its original state (`false`) makes the MaxiLoader run again, so let the editor load firsts and generates a much better and smoother UX.
4. Also simplified the way we are creating the styles saver element on `maxiBlockComponent`, that currently runs with much less amount of lines and conditionals.
5. Removed the React `< 18` `render` support as WordPress 6.2 comes with React 18 by default and is not necessary anymore 👍 

# Description

This implementation aims to fix some issues with FSE related with styling blocks and the speed. This is a first approach, we can still improve the speed a bit 👍 

Fixes #4544

# How Has This Been Tested?

This PR needs to be tested in EE and FSE:
1. Checking it loads correctly the styles after saving
6. Checking the different responsive breakpoints
7. In FSE, check it works correct with Templates and Templates Parts 👍 

Remember that changing the WP native responsive is broken #4703 👍 

# Test checklist

**_ Front/Back Testing _**

**_ Pre-Code Testing _**

# Checklist


Olekrut (approved):
Mach faster 🥇

Olekrut (approved):

Fybex (changes requested):

myroslavve (commented):
Had one comment, everything else looks good to me.

Fybex (approved):

myroslavve (approved):

elzadj (approved):

PR #4836, opened 2023-05-11
Adds `forceUseBreakpoint` feature

Naaaaiix:
# Description

This PR fixes the issue found on staging.devdemo.maxiblocks.com. To whom who wants to understand the issue, here's the situation; it's hard to reproduce but we might need to check in other helpers:
This concrete situation has been produced by a Text Maxi that has [this code editor](https://github.com/maxi-blocks/maxi-blocks/files/11452980/forUseBreakpoint.txt). This block contains next `line-height-unit` attributes:
```json
{
    "line-height-unit-general": "px",
    "line-height-unit-l": "px",
    "line-height-unit-xxl": "em",
}
```
So, what we should expect on frontend is, on a XXL screen to have `line-height: ?em` and when is not XXL to have `line-height: ?px`. When we save this block on a baseBreakpoint < XL, it saves the frontend styles correctly, but when we are on baseBreakpoint XXL, it saves the line-height CSS setting in all cases using `em`. The reason is: we modified `getLastBreakpointAttribute` to get always the baseBreakpoint attribute over the `general` attribute in case it exists. But this is just a method to deal with the EE and what we display there to be sure we offer a good UX, but is not the same for saving styles. Basically, we can't depend on the baseBreakpoint from where we are saving the styles to generate one type of style or other; the frontend styles should be the same wherever baseBreakpoint we are saving. So, for it, I've create a new parameter on `getLastBreakpointAttribute` called `forceUseBreakpoint` to avoid this mechanism that gives preference to the baseBreakpoint attribute over the `general` attribute on this function. Couple of thoughts before ending:
1. Probably we should implement this in all helpers, as they don't need to interact with baseBreakpoint and we have to ensure the same styles not depending on it. Not doing in this PR as it needs to be controlled and done with love, creating tests and ensuring a good functionality of this feature,
2. This situation IS EXTREMELY HARD TO REPRODUCE, and tbh, considering this block was created on version `0.0.1 SC1`, it's probably that with the improvements we did and the way we clean the attributes with the filters it shouldn't happen. I've tried to reproduce in different ways and I couldn't, so, what I try to share with this comment is: in case the issue is there and persists, is hard to find. Anyway, will derive into a new issue and we'll discuss about it 👍 

# How Has This Been Tested?

From `master` branch, get the code editor shared above and paste in the editor. Then, try this:
1. Using the responsiveness of the Chrome console, save from XXL baseBreakpoint. Then check frontend from XXL and <XL: it should break <XL styles giving a line-height in `em` 👍 
2. Do the same, but saving from <XL breakpoint: it should look perfect in all responsive screens.

With this branch, saving in both situations should return same result 👍 

PS: when I use the expression `<XL`, I mean lower responsive than XL 👍 

# Test checklist

**_ Front/Back Testing _**

**_ Pre-Code Testing _**

# Checklist

### DIFF
--- a/src/extensions/styles/getLastBreakpointAttribute.js
+++ b/src/extensions/styles/getLastBreakpointAttribute.js
@@ -32,7 +32,8 @@ const getLastBreakpointAttributeSingle = (
 	attributes,
 	isHover,
 	avoidXXL,
-	keys
+	keys,
+	forceUseBreakpoint = false
 ) => {
 	const { getBlockAttributes, getSelectedBlockClientId } = select(
 		'core/block-editor'
@@ -67,6 +68,7 @@ const getLastBreakpointAttributeSingle = (
 	// give priority to baseBreakpoint attribute just when the currentBreakpoint it's 'general'
 	// or the baseBreakpoint is different from 'xxl' and currentBreakpoint
 	if (
+		!forceUseBreakpoint &&
 		breakpoint === 'general' &&
 		(currentBreakpoint === 'general' ||
 			(baseBreakpoint !== 'xxl' && currentBreakpoint !== baseBreakpoint))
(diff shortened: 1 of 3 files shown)


Olekrut (approved):

Fybex (approved):

myroslavve (approved):
Couldn't reproduce in any other case, but I agree that it should be implemented for other helpers too 👍

elzadj (approved):

PR #5334, opened 2023-12-15
Issue 5333 - Search button styles in toolbar

### LINKED ISSUE #5333: Search button styles in toolbar
### Description

Link:

screenshot-wordpress local-2023 12 15-12_40_56

Also please check it for all blocks with a link.

Add a block with a link and check the toolbar.

_No response_

### URL and/or Code editor

_No response_

### How to fix it

_No response_

### Other information

_No response_

kyrapieterse:
Fixes https://github.com/maxi-blocks/maxi-blocks/issues/5333

## Summary by CodeRabbit

- **Style**
	- Updated the layout and visual design of certain UI elements for improved user experience.

### DIFF
--- a/src/components/link-control/editor.scss
+++ b/src/components/link-control/editor.scss
@@ -104,26 +104,7 @@
 		width: 280px;
 
 		&__search-submit {
-			background-color: #e5f2f8;
-			width: auto;
-			float: right;
-			font-weight: 700;
-			color: var(--maxi-secondary-color);
-			height: 37px;
-			padding: 0px 20px !important;
-			box-shadow: none !important;
-			right: -19px !important;
-			position: absolute;
-
-			svg {
-				display: none;
-			}
-
-			// WP 6.2.2 and lower support
-			&.has-icon::before {
-				content: 'Search';
-				display: block;
-			}
+			display: none;
 		}
 		&__settings {
 			border-top: 0;
@@ -186,7 +167,6 @@
 			border: 1px solid var(--maxi-grey-1-color);
 			border-radius: 0;
 			padding: 8px;
-			color: #5f6063 !important;
 			&-action {
 				background-color: var(--maxi-primary-color);
 				box-shadow: none !important;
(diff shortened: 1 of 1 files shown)


Olekrut (changes requested):
1) Changes influence to another buttons
image
2) and fix is not correct at all, button, overlap input field.
https://jam.dev/c/0007271d-adf0-4043-b050-158d03fd2384

elzadj:
yeah, I think we can just remove / hide this Search button from the url. Search works without it, when you type.

kyrapieterse:
right: -19px !important;
position: absolute;
			
is what broke the other buttons lol

beyond me 🤪 @Marko64  take over and hide it 🙏

Marko64:
Hidden the button now.

By default there's this arrow, we can have that too if  it looks better:
Screenshot_2

Olekrut (changes requested):
1) loader circle is overlapping link
image
2) sometimes link is suggesting, but not always 
I understand that I need to press Enter, but will it be intuitive for customers?
https://jam.dev/c/e834fd9f-a80c-4703-a781-c5b701ad5f14
3) is not related to this issue, but guess, easy to fix it here...
Need to change text colour for suggestion`s links ( grey or black )
image

Marko64:
1. No good way to solve this. If I add right padding for the text it looks weird without the spinner icon.
2. I think it doesn't suggest anything for you because nothing similar exists to the content you added.
3. Not able to reproduce:
Screenshot_4
, text black for me in FF and Chrome

elzadj:
1. How about just hiding that circle? Do we really need it?

Olekrut:
About 1 and 2, agree with SV)
3) The same on incognito and FF, but looks like it is only on my local. Sv does not have this bug, so ignore
image

Marko64:
2. Isn't enter reserved to confirm the input now? Especially when we don't have the button anymore. Default WP search also works like this. It shows suggestions as you type, not on enter

kyrapieterse:
MVP 🙏

elzadj (approved):

PR #5495, opened 2024-03-12
Issue 5248 - reusable blocks styles

myroslavve:
# Description

Used CSS `:has()` selector this time, since it is finally supported by all the desktop browsers.
Also removed some code that was supposed to fix problems with reusable blocks, specifically https://github.com/maxi-blocks/maxi-blocks/pull/4967#issuecomment-1701418671 and https://github.com/maxi-blocks/maxi-blocks/pull/4967#issuecomment-1702843990, but checked and seems those problems are not present.

Fixes #5248 

# How Has This Been Tested?
Checked that reusable blocks look the same as normal blocks on editor and frontend, and don't have errors.

# Test checklist

**_ Front/Back Testing _**

-   [ ] Check that reusable blocks look the same as normal blocks on editor and frontend.
-   [ ] Check https://github.com/maxi-blocks/maxi-blocks/pull/4967#issuecomment-1701418671 and https://github.com/maxi-blocks/maxi-blocks/pull/4967#issuecomment-1702843990

# Checklist

-   [x] My code follows the style guidelines of this project
-   [x] I have performed a self-review of my code
-   [x] I have commented my code, particularly in hard-to-understand areas
-   [x] My changes generate no new warnings/errors
-   [x] New and existing unit tests pass locally with my changes

## Summary by CodeRabbit

- **New Features**
	- Enhanced the `MaxiBlock` component to support full-width styles for reusable blocks.
- **Style**
	- Updated CSS for better targeting and styling of elements within the maxi-themes banner.
- **Refactor**
	- Improved attribute handling logic for blocks to consider parent blocks' attributes.
	- Simplified the `MaxiBlockComponent` by removing unnecessary logic and optimizing style handling methods.

PR #5545, opened 2024-04-10
Issue 5112 - aria labels input

### LINKED ISSUE #5112: Create a component that allow adding aria-labels
### Discussed in https://github.com/maxi-blocks/maxi-blocks/discussions/4845

<div type='discussions-op-text'>

<sup>Originally posted by **ckp267** May 12, 2023</sup>
**From Josh**
I'd like to share this with you and let you choose the best option. Kyra requested yesterday to check the score and rank Google Lighthouse was giving for a page; found out that accessibility wasn't reaching all the score we'd like, and one of the points is related with the aria-labels. Been reading about it and found a really interesting link => https://www.w3.org/WAI/ARIA/apg
I understand you aren't going to read right now, so giving a bit of resume: ARIA or  Accessible Rich Internet Application is used to generate accessible experience in the web. The first suggestion is No ARIA is better than Bad ARIA., so before we start using it we should be cautious with it. Then they have this other page which I think is incredible and will make our job easier: https://www.w3.org/WAI/ARIA/apg/patterns/. They share how to implement this aria-labels and what they need. So, here's the situation I'd like you to let me know what you prefer:
1. I can do some fast fix for the links to score correctly on the Google Lighthouse, which will be faster but will have some risks ( No ARIA is better than Bad ARIA.) and will affect just links, which means we can test another page on Google Lighthouse and say error of accessibility in other blocks.
**2. I can take a bit longer and create a component that allow adding aria-labels to creators and ensure a whole control of the accessibility to have an easier way to score 100.**

What do you think is better with the current context? Don't hesitate on asking what you need 🙏🙂

**Kyra Pieterse, Yesterday 11:28**
option 2 sounds better but lets see what C says

**Joshua Juan Caparrós, Yesterday 11:28**
cool, I don't think second option will take me longer than 2 or 3 days and will ensure a good accesibility feature 🙂

**Christiaan Pieterse (yc), 8 min**

Thank you for the info @Joshua Juan Caparrós yeah, option 2 seems like a better way to go long-term. The important thing is that it's not a quick fix - it should be included in all planning and thinking. It's easier to add it in from the beginning than to try to fix it later.  If we add it as a component then we can help make it easier for people to do it themselves. That makes sense.
More ideas here.
https://theplusblocks.com/docs/aria-labels-in-wordpress/ 
https://fullsiteediting.com/lessons/accessibility-in-full-site-editing-themes/

Joshua Juan Caparrós, 7 min

thanks for the answer </div>

myroslavve:
# Description
Added `Aria label` tab in advanced settings tab:
image

Fixes #5112 

# How Has This Been Tested?
Checked that all blocks can have aria-label for their targets by looking in dev tools inspector.
image

# Test checklist

**_ Front/Back Testing _**

-   [ ] Test the component for all blocks (aria-label is only added on frontend).

# Checklist

## Summary by CodeRabbit

- **New Features**
  - Enhanced accessibility across various components by adding `aria-label` attributes.
  - Improved handling of ARIA labels in maps, buttons, accordions, containers, dividers, and more for better user experience.
  
- **Refactor**
  - Updated components to include new ARIA label categories, improving accessibility and code maintainability.
  
- **Documentation**
  - Updated import statements and incorporated ARIA label handling functions across multiple components to align with accessibility standards.


myroslavve:
Didn't add defaults for aria labels, they need to be added on per-block basis:
When an element has an aria label, the screen readers just read the aria label and ignore the text content of the element, so if we add default aria labels it could do more harm than good, for example it would be better if screen reader read the content of the button instead of our default aria label - so it is better to leave for creators to add aria labels where they need them. Also, most blocks don't really need aria labels by default, because aria labels are intended either for interactive elements or landmarks which blocks aren't by default.
You can also read more about aria labels here: https://www.aditus.io/aria/aria-label/

Olekrut (changes requested):
Error if create blocks on master and enable this brunch
image
The same if create any blocks on this brunch
https://jam.dev/c/9250d89e-f700-4586-85bf-526078435973

myroslavve:
> Error if create blocks on master and enable this brunch image The same if create any blocks on this brunch https://jam.dev/c/9250d89e-f700-4586-85bf-526078435973

Can't reproduce this, and the issue seems to be with `loadable-components`, I suspect it is the same as described here:
https://github.com/maxi-blocks/maxi-blocks/pull/5495#issuecomment-2002044052

Olekrut (approved):
It works.
one question - should we have ability to add aria label when div is disabled?
for example icon for button or circle for NC, etc

myroslavve:
> It works. one question - should we have ability to add aria label when div is disabled? for example icon for button or circle for NC, etc

Did this, here is a list of what will be hidden:

**Button Maxi (maxi-blocks/button-maxi):**

- Icon (filtered out if icon-content is empty)

**Container Maxi (maxi-blocks/container-maxi):**

- Top shape divider (filtered out if shape-divider-top-status is false)
- Bottom shape divider (filtered out if shape-divider-bottom-status is false)

**Map Maxi (maxi-blocks/map-maxi):**

- Popup (filtered out if map-markers is empty)

**Number Counter Maxi (maxi-blocks/number-counter-maxi):**

- Circle (filtered out if number-counter-circle-status is true, note the reversal in logic)

**Search Maxi (maxi-blocks/search-maxi):**

- Close icon (filtered out if the skin is not 'icon-reveal')

**Slider Maxi (maxi-blocks/slider-maxi):**

- First arrow (filtered out if navigation-arrow-first-icon-content is empty)
- Second arrow (filtered out if navigation-arrow-second-icon-content is empty)
- Dot and All dots (filtered out if navigation-dot-icon-content is empty)

**Text Maxi (maxi-blocks/text-maxi):**

- Text (filtered out if isList is true)
- List (filtered out if isList is false)

**Video Maxi (maxi-blocks/video-maxi):**

- Play icon (filtered out if the player type is not 'popup' or play-icon-content is empty)
- Close icon (filtered out if the player type is not 'popup' or close-icon-content is empty)

elzadj (changes requested):

elzadj (approved):

PR #5720, opened 2024-09-10
Experimental: Optimization style component

elzadj:


Olekrut:
I see difference, faster 💯

PR #5757, opened 2024-10-02
Experimental: Mega menu

elzadj:
# Description

Allows to add a relation in IB on the block itself. Allows to add a Relation for Show / hide block.

Need for future mega menus. 

# How Has This Been Tested?

Here are 2 variations: with transform scale and show / hide (this one can't be animated, so maybe we should not use it). 

Transform: 
[mega-menu-transform.txt](https://github.com/user-attachments/files/17230301/mega-menu-transform.txt)

Display: 
[mega-menu-display.txt](https://github.com/user-attachments/files/17230332/mega-menu-display.txt)

Please check what we are missing, and maybe there are other ways to use IB for mega menus here. Note the IB on the block itself, and the IB settings.

Also not sure what to do for mobile. Ideas?

# Test checklist

**_ Front/Back Testing _**

**_ Pre-Code Testing _**

# Checklist

## Summary by CodeRabbit

- **New Features**
	- Introduced a new control for displaying or hiding blocks in the settings.
- **Improvements**
	- Enhanced logic for handling block interactions and effects, improving user experience.
	- Added early return conditions to optimize relation updates, preventing unnecessary processing.
	- Updated styles for better layout and responsiveness in the editor.
- **Bug Fixes**
	- Updated display styles to ensure correct rendering based on new conditions.

### DIFF
--- a/core/class-maxi-styles.php
+++ b/core/class-maxi-styles.php
@@ -1134,8 +1134,8 @@ class MaxiBlocks_Styles
             $js_var = str_replace('-', '_', $script);
             $js_var_to_pass = 'maxi' . str_replace(' ', '', ucwords(str_replace('-', ' ', $script)));
             $js_script_name = 'maxi-' . $script;
-            //$js_script_path = '//js//min//' . $js_script_name . '.min.js';
-            $js_script_path = '//js//' . $js_script_name . '.js';
+            $js_script_path = '//js//min//' . $js_script_name . '.min.js';
+            //$js_script_path = '//js//' . $js_script_name . '.js';
 
             $block_meta = $this->custom_meta($js_var, false);
             $template_meta = $this->custom_meta($js_var, true);
--- a/core/class-maxi-styles.php
+++ b/core/class-maxi-styles.php
@@ -1134,8 +1134,8 @@ class MaxiBlocks_Styles
             $js_var = str_replace('-', '_', $script);
             $js_var_to_pass = 'maxi' . str_replace(' ', '', ucwords(str_replace('-', ' ', $script)));
             $js_script_name = 'maxi-' . $script;
-            $js_script_path = '//js//min//' . $js_script_name . '.min.js';
-            //$js_script_path = '//js//' . $js_script_name . '.js';
+            //$js_script_path = '//js//min//' . $js_script_name . '.min.js';
+            $js_script_path = '//js//' . $js_script_name . '.js';
 
             $block_meta = $this->custom_meta($js_var, false);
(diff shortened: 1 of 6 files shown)


Olekrut (changes requested):
1)  https://jam.dev/c/ce065c26-42f8-4074-9d19-10788b82591e
 I understand why the scroll is appearing on the page, and maybe if the page is longer, everything will be fine, but if the mega menu container is longer than the page, it looks not so good
2) https://jam.dev/c/a6525065-b0ee-49bb-934c-c36c6a216f6b
you remade bit triggers for continuous hover, but it doesn't work so well with Pricing.
And I guess we need to improve the click action (when the first click is enabling IB, the second one is disabling)

3) about second variant ( display ). it works much worse and backend is not good, I don't think we should use it.

4) about mobile - yes, it is a problem
for mobile, we should use the click action, and we should understand how the mobile design looks like from start of developing (it should not be too long).

5) and need to fix backend, it because of the native group, maybe we need to allow moving a container to our group?
image

elzadj:
@Olekrut 

> 1. https://jam.dev/c/ce065c26-42f8-4074-9d19-10788b82591e
>    I understand why the scroll is appearing on the page, and maybe if the page is longer, everything will be fine, but if the mega menu container is longer than the page, it looks not so good

There is no solution here aside of a design one. 

> 2. https://jam.dev/c/a6525065-b0ee-49bb-934c-c36c6a216f6b
>    you remade bit triggers for continuous hover, but it doesn't work so well with Pricing.
>    And I guess we need to improve the click action (when the first click is enabling IB, the second one is disabling)

It's not about pricing, there is a 3px border around the menu container, need to remove it, or compensate with negative margin, so the mega item overlap it on hover.

> the first click is enabling IB, the second one is disabling

Mega menus should work on hover. Our hover already works as click on mobile.

> 3. about second variant ( display ). it works much worse and backend is not good, I don't think we should use it.

ok, no display, let's use transform. Could you also try with width / height, maybe? Opacity is not going to work on itself, so we can't use it. Any other ideas what we can use for IB here?

> 4. about mobile - yes, it is a problem
>    for mobile, we should use the click action, and we should understand how the mobile design looks like from start of developing (it should not be too long).

Asked @kyrapieterse for some mobile designs, possibly need to try to create 1-2 mega menus and try, for tests. Please use hover for now. We can extend the click later, if needed.

> 5. and need to fix backend, it because of the native group, maybe we need to allow moving a container to our group?

Fixed in that branch with css, check pls.

Olekrut (approved):
Cool) 
lets merge and need to try to make smth
@elzadj

PR #5831, opened 2024-12-19
Issue 5825 - Add "Keep only text content" switch

### LINKED ISSUE #5825: DC. Character limit 0 bug on frontend
### Description

1. Brief description
if to set 0 for dc text ( field - content ) > text have wrong colour and extra div above on frontend

2. What was expected to happen
DC text with correct colour and without extra div above

3. What actually happened
text with bug on frontend

1) you should have post with content
2) create row
3) enable CL
4) create in column text block
5) enable DC
6) choose content
7) set 0 for Character limit
8) check frontend

https://jam.dev/c/945aea03-e4b8-4790-af51-c2c8e5c34ec9

https://jam.dev/c/067097f7-715b-4f0b-b326-5a46fda355f2

### URL and/or Code editor

[dc linit 0 bug.txt](https://github.com/user-attachments/files/18119693/dc.linit.0.bug.txt)

### Other information

_No response_

myroslavve:
# Description
Added the switch, to choose whether to copy the content with tags from the post or strip all tags and leave only text.
Set to `false` by default, so that old patterns stay the same.

Fixes #5825 

# How Has This Been Tested?
Checked that with the switch enabled, text has correct styles. When disabled, there should be all content from the post, like images SVGs.

# Test checklist

**_Front/Back Testing_**

-   [ ] Checked that with the switch enabled, text has correct styles
-   [ ] When disabled, there should be all content from the post, like images SVGs

## Summary by CodeRabbit

- **New Features**
	- Introduced a toggle option for users to limit content to only text when no character limit is applied.
	- Added a new parameter to enhance content management capabilities.

- **Bug Fixes**
	- Improved logic for handling post content to ensure proper formatting and stripping of HTML tags.

- **Documentation**
	- Updated comments for clarity in the codebase.

### DIFF
--- a/core/class-maxi-dynamic-content.php
+++ b/core/class-maxi-dynamic-content.php
@@ -1931,6 +1931,7 @@ class MaxiBlocks_DynamicContent
             'dc-post-taxonomy-links-status' => $dc_post_taxonomy_links_status,
             'dc-link-status' => $dc_link_status,
             'dc-accumulator' => $dc_accumulator,
+            'dc-keep-only-text-content' => $dc_keep_only_text_content,
         ] = $attributes;
 
         $post = $this->get_post($attributes);
@@ -1979,7 +1980,7 @@ class MaxiBlocks_DynamicContent
             }
 
             // Limit content
-            if (isset($dc_limit) && $dc_limit > 0) {
+            if (isset($dc_limit) && ($dc_limit > 0 || $dc_keep_only_text_content)) {
                 $post_data = wp_strip_all_tags($post_data);
 
                 // Ensures no double or more line breaks
--- a/src/components/dynamic-content/index.js
+++ b/src/components/dynamic-content/index.js
@@ -723,7 +723,7 @@ const DynamicContent = props => {
 										{limit === 0 && field === 'content' && (
 											<ToggleSwitch
 												label={__(
-													'Keep only text content',
+													'Display as plain text',
 													'maxi-blocks'
 												)}
 												selected={keepOnlyTextContent}
--- a/src/components/dynamic-content/index.js
+++ b/src/components/dynamic-content/index.js
@@ -140,6 +140,7 @@ const DynamicContent = props => {
 		customFormat,
 		acfGroup,
 		mediaSize,
+		keepOnlyTextContent,
 	} = dcValues;
 
 	const dcValuesForDate = {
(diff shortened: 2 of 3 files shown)


Olekrut (changes requested):
https://jam.dev/c/c5fcbb3f-88d5-4797-9cf0-8ca91c5663dc

it is fixed, but in this case headers does not look correct

in general, if to add not text blocks to post ( button, group, search, etc. ) They are not displaying correct on pages by DC content ( but FSE is OK )

myroslavve:
> https://jam.dev/c/c5fcbb3f-88d5-4797-9cf0-8ca91c5663dc
> 
> it is fixed, but in this case headers does not look correct
> 
> in general, if to add not text blocks to post ( button, group, search, etc. ) They are not displaying correct on pages by DC content ( but FSE is OK )

OK, maybe "Keep only text content" is not descriptive enough. This is expected to happen, we keep only text and strip all tags and styles.

Do you think that a more specific label, like: "Keep only text content (without styles)", would explain what it does better to you? 
Or do you think that current label is ok to leave?

Olekrut:
> > https://jam.dev/c/c5fcbb3f-88d5-4797-9cf0-8ca91c5663dc
> > it is fixed, but in this case headers does not look correct
> > in general, if to add not text blocks to post ( button, group, search, etc. ) They are not displaying correct on pages by DC content ( but FSE is OK )
> 
> OK, maybe "Keep only text content" is not descriptive enough. This is expected to happen, we keep only text and strip all tags and styles.
> 
> Do you think that a more specific label, like: "Keep only text content (without styles)", would explain what it does better to you? Or do you think that current label is ok to leave?

Olekrut (changes requested):
1) label for this toggle is long already as for me, so idk, better to ask @ckp267 
2) when toggle is disabled, content should display correct or need to create another issue for it? ( is not important, because styles are not correct only for unexpected in post blocks ( search, group, button, etc) Image, SVG, text, headers look correct )
t

myroslavve:
> 1. label for this toggle is long already as for me, so idk, better to ask @ckp267

Ok, @ckp267, the toggle is for dynamic content, and basically does this: https://github.com/maxi-blocks/maxi-blocks/pull/5580#issue-2283496923, if the toggle is enabled we keep only text content with text-maxi styles, otherwise we keep the post content as it is with all images, SVGs.

The toggle is called "Keep only text content" currently, maybe you can suggest a better label?

> 2. when toggle is disabled, content should display correct or need to create another issue for it? ( is not important, because styles are not correct only for unexpected in post blocks ( search, group, button, etc) Image, SVG, text, headers look correct )

Let's do this in a separate issue.

ckp267:
@myroslavve @Olekrut 

After some discussion and consulting the Oracle, I suggest change label "Keep only text content"

NEW LABEL: **"Display as plain text"**

This label:

- Clearly communicates the end result
- Uses familiar terminology
- Implies the removal of formatting without being too technical
- Works well in the context of the Dynamic Content UI panel
- Aligns with WordPress conventions

Unless we are missing something else completely - no further discussion needed. You can make the change and proceed.

Olekrut:
@myroslavve  @ckp267 what if we add an icon for information like for the date and when hovering over it we show a more detailed answer for uneducated, in this question, users?

ckp267:
> @myroslavve @ckp267 Cool, but what if we add an info icon like for the date and when hovering over it we show a more detailed answer for uneducated, in this question, users?

Let's wait for incoming support requests before we add more. It's good to go now.

@Olekrut @myroslavve 

Update and merge please.

Olekrut (approved):

elzadj (approved):

PR #5839, opened 2024-12-24
Issues 4343, 3699 - fixed the helper errors

### LINKED ISSUE #4343: Size helper error
### Description

Seems we've got a situation with Size helper

1. Create a block
2. Change the unit
3. Add and remove a value

Captura de Pantalla 2023-02-09 a las 20 26 28

### URL and/or Code editor

_No response_

### How to fix it

Please, add a unit test for it 👍 

### Other information

_No response_

myroslavve:
# Description

This PR fixes two issues with the same problem in different components.

Fixes #4343 
Fixes #3699

# How Has This Been Tested?
Repeated steps from #4343, also checked for width. For background layers, did the same steps as for height, and checked the styles.

# Test checklist

**_Front/Back Testing_**

-   [ ] Check that width and height don't have the problem anymore
-   [ ] Check that background layers position doesn't have the problem anymore

## Summary by CodeRabbit

- **New Features**
	- Introduced a new utility function to validate numeric values, enhancing various style-related functionalities.

- **Bug Fixes**
	- Enhanced validation in size styles logic to prevent inclusion of invalid values.
	- Improved handling of position values in background styles to ensure valid attributes are processed correctly.
	- Refined logic for margin and padding styles to ensure only valid numbers are considered.
	- Updated position styles logic to ensure only valid numeric values are processed.
	- Improved SVG styles logic to ensure only valid numeric values for icon size are considered.

- **Tests**
	- Added a new test case to verify the function returns an empty object when width/height is empty and unit is not empty.
	- Introduced a new test case to check the behavior of background styles when position values are empty strings while units are specified.


Olekrut (changes requested):
it works for new blocks, but not for old
[helper code.txt](https://github.com/user-attachments/files/18262627/helper.code.txt)
test code from master  Better to test it on Image block)
Position, margin, padding, height, width, (min, max as well ) and canvas

Should i test it for all blocks or it will be fixed for all in one time?

myroslavve:
> it works for new blocks, but not for old [helper code.txt](https://github.com/user-attachments/files/18262627/helper.code.txt) test code from master Better to test it on Image block) Position, margin, padding, height, width, (min, max as well ) and canvas
> 
> Should i test it for all blocks or it will be fixed for all in one time?

Fixed for all settings. It will work the same for all blocks. And also, should work for old blocks as well, worked for me when I tested.

Olekrut (changes requested):
https://jam.dev/c/ee60d60b-17de-4d5d-8829-287fd5e16830
looks ok, but for old blocks > only after resave of page
@elzadj surmises that need migrator

myroslavve:
> https://jam.dev/c/ee60d60b-17de-4d5d-8829-287fd5e16830 looks ok, but for old blocks > only after resave of page

If the problem is on frontend only, it is ok. We cannot change frontend pages without user action, so the styles should remain the same until user updates a page.

Olekrut:
> If the problem is on frontend only, it is ok. We cannot change frontend pages without user action, so the styles should remain the same until user updates a page.

but we are talking about errors, so it will not make any visual changes, no?

myroslavve:
> > If the problem is on frontend only, it is ok. We cannot change frontend pages without user action, so the styles should remain the same until user updates a page.
> 
> but we are talking about errors, so it will not make any visual changes, no?

Yes that's right. It just seems to me like this problem isn't that big, but we will have to create a whole new system to update frontend pages, so isn't worth it IMO.

Olekrut (approved):

elzadj (approved):

PR #5876, opened 2025-02-04
Add import alias auto replace

myroslavve:
# Description

Since we added aliases, I have been writing imports that end up not using aliases, when they should, because they are written with `..` by default by AI and VS Code IntelliSense, and I usually don't notice it.
So I added ESLint plugin that marks those imports as errors and replaces them with correct ones when file is saved.

# How Has This Been Tested?
Checked that imports with `..` get replaced. Also made sure that sibling level imports like `./index.js`, are not overwritten with aliases.

## Summary by CodeRabbit

- **Chores**
  - Enhanced configuration for consistent module import alias management.
  - Added a new development dependency to support advanced module handling.


elzadj (changes requested):
Doesn't seem to be working for me :thinking: 
Please re-check the configuration, and maybe make a video with a working example? Thanks.

myroslavve:
I just copied some basic configuration from the plugin's documentation, so I don't think it's the problem (works for me).

I suppose the only problem could be that the package wasn't properly installed for you,
so I just did `npm ci --legacy-peer-deps` and everything works for me here's a video:

https://github.com/user-attachments/assets/d4508655-a11f-42e5-ac56-61c8d7a31893

elzadj:
@myroslavve got it, it works if I open the maxi-blocks folder in the editor, not the wp root folder.

Please, fix conflicts, and it's ready to be merged, thanks.

elzadj (approved):

PR #5958, opened 2025-05-21
Issue 4922 - fix the validation interfering with user typing by debouncing it

myroslavve:
# Description

Fixes #4922 

# How Has This Been Tested?
Checked that I can type 33 in map marker size input. Checked that validation still works, tried typing really fast, really slow, the validation should happen only 300ms after you stopped typing, so you can type 1000000 and it will validate only after you stop. The styles shouldn't be changing before validation.

# Test checklist

**_Front/Back Testing_**

-   [ ] Should work in all number inputs with min/max value, so check not only map marker
-   [ ] Let me know if validation applying with delay after you typed feels weird

# Checklist

## Summary by CodeRabbit

- **Refactor**
  - Improved input handling for number controls to provide more consistent and responsive updates when changing values.

### DIFF
--- a/src/components/advanced-number-control/index.js
+++ b/src/components/advanced-number-control/index.js
@@ -206,7 +206,13 @@ const AdvancedNumberControl = props => {
 					onChangeValue(val);
 				}
 			},
-			[onChangeValue, optionType]
+			// onChangeValue is intentionally omitted from dependencies to prevent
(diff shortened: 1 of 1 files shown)


Olekrut (approved):
is not ideal, but we can work with it at least. Good enough
https://jam.dev/c/4c334a0d-1515-42dc-8ae5-09505ae66038

PR #6042, opened 2025-10-03
Add customizable popup animation duration to video block

### LINKED ISSUE #3793: Video. Need to add transition for Zoom in/out
image
Colour, opacity, etc) and maybe add to custom CSS

kyrapieterse:
Fixes https://github.com/maxi-blocks/maxi-blocks/issues/3793

## Summary by CodeRabbit

* **New Features**
  * Adjustable popup animation duration (value + unit) for the Video Maxi block.
  * New settings control to configure and reset popup animation timing.

* **Improvements**
  * Popup animation reliably replays when a popup opens and respects configured timing while preserving final state.
  * Form controls no longer leak private props and toolbar inputs use per-instance IDs for proper label association.
  * Editor icons now have consistent sizing.

* **Tests**
  * E2E formatting tweaks for stability.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


Olekrut (changes requested):
fixed. but

1) error, broken block. Is need migrator?
[video.bug.txt](https://github.com/user-attachments/files/24494791/video.bug.txt)
2) Unit is unneeded, I guess, better to remove it at all

image
Should I create for it new issue?

Olekrut (changes requested):
Actually, good, one more thing. When we chose Zoom IN/OUT
transition field is empty, but we have there 0.3s already, so maybe need to add it as placeholder?
And maybe need to add Easing dropdown there?
image

PR #6104, opened 2025-12-16
Adjust flex wrapping in color control editor styles

### LINKED ISSUE #6102: search bugs
- [ ] #6100
Image

kyrapieterse:
Added flex-wrap: wrap to .maxi-color-control__palette-container and set flex-wrap: nowrap for &-container to improve layout handling in the color control editor component.

fixes https://github.com/maxi-blocks/maxi-blocks/issues/6102


Olekrut (changes requested):
It is fixed but instead of find the reason of problem, you are fixing bug in another way, and add extra styles instead of change existed.

You have added wrap to nowrap here
 
image

and theni it is  overwriting -  nowrap to wrap here for custom palette 

image

How need to fix it - need to find reason of problem - here ( padding ) : 
https://jam.dev/c/90a53e23-68fa-4949-84c7-ec4c78487b6f

More interesting, why here is another padding for the same component, probably would be better just add this class to other style, where padding 0 8px; so everywhere will be the same style in one place, so we can even optimize a bit with help it

P.S.: Am i wrong?

PR #6128, opened 2025-12-20
Refactor CSS variable names for consistency

kyrapieterse:
An attempt at fixing the speed


kyrapieterse:
Notes for @elzadj 

I was unable to rename maxi-blocks--active, maxi-container-block maybe there is a way you can. extremely heavy these 2

PR #6143, opened 2026-01-06
Lazy-load map editor and skip Leaflet load when no maps

kyrapieterse:
### Motivation
- Map editor code imports `react-leaflet`/`leaflet` which increases the base editor bundle size even when no map block is present. 
- The frontend runtime `js/maxi-map.js` unconditionally injected Leaflet assets, risking duplicate or unnecessary downloads when no maps exist.

### Description
- Dynamically import the editor map UI by replacing the direct `MapContent` import with `lazy()` and wrapping usage in `Suspense` in `src/blocks/map-maxi/edit.js` so Leaflet-related code is split from the main editor bundle.  
- Add an early-return guard in `js/maxi-map.js` to skip loading Leaflet CSS/JS and Google Mutant scripts when `mapItems.length === 0`.  
- Update imports to use `lazy` and `Suspense` from `@wordpress/element` and keep existing behavior identical when a map is present.

### Testing
- No automated tests were executed for this change.

------
[Codex Task](https://chatgpt.com/codex/tasks/task_e_695c9fef872483219f9485cd87c4ac1a)

## Summary by CodeRabbit

* **Performance Improvements**
  * Maps now load lazily and skip work when no map data exists, reducing initial load and unnecessary processing.
  * External map script loading is centralized and reused to prevent duplicate requests.

* **Bug Fixes / Reliability**
  * Improved cleanup and cancellation to avoid leftover map layers and race conditions during unmounting.
  * Added user-facing error handling for map load failures.


Olekrut (changes requested):
Google maps Api is not displaying on backend

Image

Olekrut (approved):

PR #6191, opened 2026-02-17
Fix AdvancedNumberControl inline meta for spinner, reset, and unit-cl…

kyrapieterse:
## Description

Fixes `AdvancedNumberControl` inline unit metadata handling without letting unit-only changes rewrite the current numeric value.

The failing `e2e-js` run showed snapshot drift across background, relation, and typography controls, plus an AxisControl regression where `auto` was persisted as `NaN`. The root cause was the unit selector path calling `onChangeValue` and scheduling the debounced `handleChange` even when only the selected unit changed. Since the debounced callback reads `latestValueRef.current` later, it could persist stale or converted values after the UI state had changed.

## Changes

- Keep unit-only selector changes as unit updates only.
- Preserve inline unit metadata for value changes that need an inline preview.
- When switching units requires clamping, update the inline preview with the selected unit and persist the clamped value directly.
- Avoid the delayed ref-based persistence path for unit selector changes, preventing `auto` from becoming `NaN`.
- Keep optional callback guards in place.

## Testing

- `npm run build`
- `git diff --check`
- GitHub Actions on `effda7da2`: `unit-js` passed.
- GitHub Actions on `effda7da2`: `Analyze (javascript-typescript)` passed.

Targeted local e2e was attempted with:

```bash
npm run test:e2e -- e2e-tests/components/axis-control/index.spec.js --runInBand
```

That local run could not complete because the WordPress e2e harness timed out in `beforeAll` and the REST nonce request returned `400` from `localhost:8889`. The pushed GitHub e2e jobs should be used for final validation.


kyrapieterse:
Checked the failing CI for this PR.

Current head: `3635a628b3cd84629b3a391618521f849218aa92`

Status:
- `unit-js`: passing
- `e2e-js-playwright`: passing
- `Analyze (javascript-typescript)`: passing
- `e2e-js`: failing in the `Running the tests` step

Failing job: https://github.com/maxi-blocks/maxi-blocks/actions/runs/22673400557/job/65723283649

What is failing:
- Background layer hover/responsive snapshots are drifting in `shapeLayer`, `colorLayer`, `videoLayer`, and `imageLayer`.
- `AxisControl auto` persists `NaN` instead of `auto` for margin values.
- Relation builder button icon snapshots show units reverting from `%` to `px` for icon width/padding.
- Typography control has one assertion failure after the same e2e run.

Likely root cause:
`src/components/advanced-number-control/index.js` now calls `onChangeValue` and schedules `handleChange` inside the unit selector `onChange` path even when the value does not need clamping. That means a unit-only change also creates extra value writes. Because the debounced `handleChange` reads `latestValueRef.current` later, it can persist a different value than the one present when the unit was selected. This matches the observed failures:
- unit selections can be overwritten/reverted to the previous/default unit (`px` instead of `%`), causing snapshot drift;
- if the value becomes `auto` before the debounced callback fires, `+latestValueRef.current` becomes `NaN`, which is exactly what the AxisControl failure reports.

Recommended fix:
- Do not call `onChangeValue`/`handleChange` for non-clamping unit changes; keep that path as a pure `onChangeUnit?.(val)` update.
- For the clamping branch, persist the clamped value directly and avoid relying on a delayed callback that rereads the mutable ref.
- Add/adjust coverage for unit selection after typed values and AxisControl `auto` so this does not regress again.

kyrapieterse:
Pushed a fix in `effda7da205e7b3598d1de2721e6e6579f877788`.

Change made:
- `AdvancedNumberControl` no longer emits value writes for unit-only changes.
- If switching units requires clamping, it still updates the inline preview with the new unit and then persists the clamped value directly, without the delayed ref-based `handleChange` call that could later read `auto` and write `NaN`.

Local verification:
- `npm run build` passed.
- `git diff --check` passed.
- `npm run lint:js -- src/components/advanced-number-control/index.js` is blocked by the repo's current ESLint 9 / missing `eslint.config.*` setup.
- Local targeted `npm run test:e2e -- e2e-tests/components/axis-control/index.spec.js --runInBand` could not run cleanly because the local WP e2e harness timed out in `beforeAll` and returned `400` for the REST nonce request.

GitHub status at the time of this update:
- `unit-js`: passed
- `Analyze (javascript-typescript)`: passed
- `e2e-js` / `e2e-js-playwright`: still running

PR #6217, opened 2026-03-10
Wp 7 beta 5 fixes

elzadj:
Fixes https://github.com/maxi-blocks/maxi-blocks/issues/6215

## Summary by CodeRabbit

* **Bug Fixes**
  * Tightened editor CSS scope to the editor wrapper to prevent style leakage and duplicates.
  * Adjusted asset enqueuing to avoid duplicate styles in admin and frontend.

* **Chores**
  * Updated block registration to WordPress block API version 3.
  * Made custom script registrations conditional behind a feature toggle.
  * Filtered localized scripts based on active template/content to avoid stale payloads.
  * Improved admin dashboard typography for form controls and buttons.

* **Tests**
  * Updated end-to-end tests to operate inside the editor iframe/canvas for more reliable DOM interactions.

PR #6250, opened 2026-04-16
2.1.10

elzadj:
2.1.10, tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Security fix for authenticated stored XSS in Style Card API (CVE-2026-6378).
  * Require stronger permission to remove custom image sizes.
  * Sanitize stored and inline style values to block unsafe content.
  * Fix Interaction Builder relations disappearing when multiple relations target the same block.
  * Improve unique-ID handling during pattern import/paste and add guard for map initialization.

* **Chores**
  * Bumped plugin/package version and metadata to 2.1.10.

* **Tests**
  * Updated e2e snapshots to 2.1.10 and improved preview/test synchronization (non-fatal waits).


Olekrut:
https://github.com/maxi-blocks/maxi-blocks/pull/6217#issuecomment-4127171945
the same was here, only local

[Map error 1.txt](https://github.com/user-attachments/files/26967156/Map.error.1.txt)

image


Else looks ok

PR #6257, opened 2026-05-07
Optimize breakpoint style cache updates

kyrapieterse:
## Summary

This PR improves editor responsiveness when switching MaxiBlocks breakpoints and fixes the stale breakpoint CSS cache path that could make blocks look broken after responsive changes.

The key behavior change is that breakpoint-only renders now reuse cached CSS where possible instead of forcing every selected block through full style generation. When block styles do regenerate, the CSS cache entry for that block is replaced so future breakpoint switches cannot reuse old CSS.

## Root cause

The performance issue came from breakpoint switches causing expensive style work across many blocks. The logs showed repeated `maxiBlockComponent getStylesObject` and `displayStyles` work during responsive changes, especially for rows, columns, buttons, images, and text blocks.

The correctness bug was in `SAVE_CSS_CACHE`: once a block had a cache entry for its `uniqueID`, later regenerated styles were ignored. The active breakpoint could look correct because fresh CSS was injected immediately, but switching breakpoints could pull stale CSS from the old cache entry.

## What changed

- Added a shared MaxiBlocks profiler helper for timing and render-count logs.
- Added profiling around breakpoint resizing, `setScreenSize`, style resolving, style generation, CSS cache writes, block rendering, and inserter rendering.
- Changed CSS cache generation to use lazy per-breakpoint style getters, so CSS for a breakpoint is generated only when needed.
- Changed `SAVE_CSS_CACHE` to replace an existing cache entry when styles regenerate, preventing stale breakpoint CSS.
- Preserved lazy cache descriptors when raw CSS is merged into an existing cache entry.
- Added a cache selector path that supports the `MemoCache`/LRU cache instance.
- Updated `MaxiBlockComponent` so pure breakpoint changes can inject cached breakpoint CSS directly and update responsive classes without full style generation.
- Added viewport-unit detection so blocks using viewport units still regenerate styles when breakpoint context matters.
- Memoized inserter-related components and reduced selected block render churn.
- Simplified the loader check to inspect the current block by `clientId` instead of traversing all blocks.
- Added selected-block responsive audit logs for debugging breakpoint attributes and computed display snapshots.
- Updated the reducer test so regenerated styles must replace existing CSS cache entries.

## User impact

Breakpoint switching should feel much faster in larger layouts while still rendering the correct responsive CSS after edits. The previous stale-cache case should no longer leave blocks visually out of date or broken after changing styles and switching between breakpoints.

## How to test

1. Check out this branch and install dependencies if needed.
2. Run `npm test -- src/extensions/styles/store/test/reducer.js --runInBand`.
3. Run `npm run build`.
4. Load the MaxiBlocks plugin build in a local WordPress editor.
5. Open a page or pattern with nested MaxiBlocks content, especially container, row, column, image, SVG icon, text, and button blocks.
6. Switch between `general`, `xxl`, `xl`, `l`, `m`, `s`, and `xs` breakpoints.
7. Confirm the layout updates immediately and does not show unsupported-block placeholders.
8. Select a block with responsive settings, change a visible style at one breakpoint, then switch away and back. Good examples are SVG/icon width, row gap, column sizing, container width, image width, or absolute position values.
9. Confirm the latest value is preserved visually after switching breakpoints and no old/stale CSS reappears.
10. Repeat with several selected block types and nested rows/columns.
11. Watch the browser console for `MaxiBlocks perf (1s)` logs. Breakpoint-only switches should show cached CSS paths where available and avoid large repeated full-style generation across every block.
12. Confirm there is no `Module build failed` error for `html-dom-parser`/`domhandler` after rebuilding and hard-refreshing the editor.

## Validation

- `npm test -- src/extensions/styles/store/test/reducer.js --runInBand` passed.
- `npm run build` passed with the existing asset-size warnings.

## Summary by CodeRabbit

* **New Features**
  * Added an opt-in performance profiler and a responsive attribute audit/logging tool.

* **Performance Improvements**
  * Reduced editor re-renders via memoization of inserters and related UI.
  * Faster, cache-first responsive style handling and improved CSS cache accounting.

* **Bug Fixes**
  * Simplified maxi-block loader detection for more reliable loading behavior.

* **Tests**
  * Updated selector and style-cache tests; normalized line endings in post-content retrieval.

[Review Change Stack](https://app.coderabbit.ai/change-stack/maxi-blocks/maxi-blocks/pull/6257)

### DIFF
--- a/src/components/block-inserter/index.js
+++ b/src/components/block-inserter/index.js
@@ -4,7 +4,13 @@
 import { __ } from '@wordpress/i18n';
 import { ButtonBlockAppender, Inserter } from '@wordpress/block-editor';
 import { select, useDispatch, useSelect } from '@wordpress/data';
-import { forwardRef, useContext, useRef, useState } from '@wordpress/element';
+import {
+	forwardRef,
+	memo,
+	useContext,
+	useRef,
+	useState,
+} from '@wordpress/element';
 import { Tooltip } from '@wordpress/components';
 
 /**
@@ -18,6 +24,7 @@ import Button from '@components/button';
 import Dropdown from '@components/dropdown';
 import Popover from '@components/popover';
 import RepeaterContext from '@blocks/row-maxi/repeaterContext';
+import { countProfile } from '@extensions/performance/profiler';
 
 /**
  * Styles
@@ -33,7 +40,7 @@ const WRAPPER_BLOCKS = [
 	'maxi-blocks/slide-maxi',
 ];
 
-const BlockInserter = props => {
+const BlockInserter = memo(props => {
 	const { className, clientId } = props;
 
 	const { selectBlock } = useDispatch('core/block-editor');
@@ -60,9 +67,9 @@ const BlockInserter = props => {
 			/>
 		</div>
 	);
-};
+});
 
-const ButtonInserter = props => {
+const ButtonInserter = memo(props => {
 	const { onToggle, style = {} } = props;
 
 	const repeaterStatus = useContext(RepeaterContext)?.repeaterStatus;
@@ -99,9 +106,9 @@ const ButtonInserter = props => {
 			</div>
 		</Tooltip>
 	);
-};
+});
 
-const WrapperBlockInserter = forwardRef((props, ref) => {
+const WrapperBlockInserter = memo(forwardRef((props, ref) => {
 	const { clientId, isSelected, hasSelectedChild } = props;
 
 	const { getBlockName, getBlockParents } = select('core/block-editor');
@@ -211,9 +218,11 @@ const WrapperBlockInserter = forwardRef((props, ref) => {
 		);
 
 	return null;
-});
+}));
 
-const InterBlockToggle = props => {
+const InterBlockToggle = memo(props => {
+	countProfile('InterBlockToggle render');
+
 	const { clientId, onToggleInserter, blockRef, isOpen } = props;
 
 	const [isHovered, setHovered] = useState(false);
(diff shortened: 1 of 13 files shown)


kyrapieterse:
Addressed the review feedback in `994a478d3`.

Changes made:
- Removed implicit default profiling on editor load.
- Removed unconditional profile boosting from breakpoint switches.
- Gated responsive audit logs behind explicit responsive-debug flags/storage keys instead of the profiler flag.
- Stripped string previews from responsive audit output so logs only include string lengths.
- Normalized cached CSS before applying the breakpoint fast path.
- Made `SAVE_RAW_CSS_CACHE` replace lazy getter descriptors with concrete data properties.
- Reused the lazy-safe cache size estimator in delete/cleanup paths.
- Tightened the viewport-unit regex to numeric viewport values.
- Added tests for cache replacement content, raw CSS replacing lazy breakpoint getters, and MemoCache selector behavior.

Skipped:
- The `isChild={!!blockRootClientId}` suggestion was not changed. In this PR's actual diff, the previous `withMaxiProps` logic used `getBlockParents(clientId)` and treated any parent as child context, so `!!blockRootClientId` preserves that behavior rather than changing it to Maxi-parent-only semantics.

Validation:
- `npm test -- src/extensions/styles/store/test/reducer.js src/extensions/styles/store/test/selectors.js --runInBand` passed.
- `npm run build` passed with existing asset-size warnings.
- `npm run lint:js -- ...` is currently blocked by the repo ESLint v9 config migration issue (`eslint.config.js` missing), not by these changes.

PR #6333, opened 2026-06-09
Sync Light/Dark typography & navigation settings in the Style Card editor

elzadj:
# SUMMARY

In the Style Card (SC) editor, Light and Dark tone globals are edited on separate tabs, but their typography is almost always identical – only colours intentionally differ. This forced users to edit typography twice. 

This PR keeps the two tones in sync by default and lets users opt out per element on the Dark tab. It is implemented entirely in the editor UI – the SC attribute/data model is unchanged. "Synced" is derived by comparing the two tones, combined with the transient (session-only) toggle state; nothing new is persisted to the style card.

# WHAT CHANGED

Typography sync: Editing typography on the Light tab mirrors the change to Dark while they are in sync. Each element on the Dark tab gets an "Override light tone typography" toggle. While synced (toggle off), the typography controls are hidden; turning the override on reveals them and lets Dark be edited independently. Override toggles are per element and per heading level (H1-H6 each). Re-syncing (toggle back off) copies Light's typography onto Dark, preserving each tone's own colours. Colours/palettes are never synced.

Navigation: Item padding is per-tone, so it is synced/overridable alongside typography (and now rendered right after the typography controls). The Dark toggle is labelled "Override light tone settings". The mobile-menu options (Overwrite mobile breakpoint, Show mobile menu for all sizes, ...down from px) and Remove item underline on hover apply globally (their CSS is not tone-scoped), so they now live on the Light tab only and are hidden from Dark.

Reset behaviour on Dark: Resetting a typography setting on the Dark tab now falls back to the current Light value instead of the factory default (e.g., Light 35 / Dark 40 -> reset Dark -> 35). This covers font-family too (the font-family reset previously cleared it to empty).

Font-weight fallback fix (bundled in this branch): When a requested font weight is not available for a font, fall back to an available bundled weight instead of assuming 400 (which some fonts do not ship). Applied on both the PHP side (class-maxi-styles.php) and the editor JS (loadFonts.js), with tests.

# KEY FILES

- src/editor/style-cards/typographySync.js (new) - derives sync state, syncable-key rules (excludes colours and global nav keys), and the Light reset-value resolver. 
- src/editor/style-cards/maxiStyleCardsEditor.js - transient override state, key-based mirroring in onChangeValue, re-sync copy, Dark reset resolver. 
- src/editor/style-cards/maxiStyleCardsTab.js - override toggle, hidden-until-overridden controls, padding placement, Light-only global nav options. 
- src/components/typography-control/index.js - optional getSCResetValue consulted by getDefault, plus defaultValue on the font-family selector. 
- core/class-maxi-styles.php and src/extensions/text/fonts/loadFonts.js (+ tests) - font-weight fallback.

# HOW SYNC STATE IS DERIVED (NO SCHEMA CHANGE)

For each element, the syncable settings of both tones are compared (typography for text elements; typography + padding for navigation, excluding colors and global keys). Equal means synced. A transient per-session map handles the one ambiguous case (user toggles override before editing). Consequence: if Dark is left identical to Light, reopening the editor shows it as synced again - which is correct, since identical is synced.

# TESTING

1. npm start, open the Style Card editor. 
2. Default: a fresh card shows Dark override toggles off with typography hidden. 
3. Mirror: change a Light typography value, enable the Dark override, and confirm Dark shows the same value (colours still per tone). 
4. Override: edit Dark independently; Light is unaffected. 
5. Re-sync: toggle override off, Dark typography snaps back to Light, colours preserved. 
6. Reset: on Dark (overridden), reset font size / font family, and fall back to the Light value. 
7. Navigation: padding behaves like typography under the toggle; the mobile/hover options appear on Light only. 
8. Per heading level: overriding H2 does not affect H1/H3. 
9. Save & reload, no SC schema/attribute changes in the saved JSON.

# NOTES / FOLLOW-UPS

Navigation padding reset still goes to the factory default (it uses PaddingControl/AxisControl, a separate reset path) - reset-to-light there would be a follow-up.

Also added a fix for fonts without normal (400) weight not loading on frontend.

## Summary by CodeRabbit

* **New Features**
  * Typography sync between Light and Dark style cards with per-element override and re-sync controls

* **Improvements**
  * Smarter font weight/style normalization and more reliable fallback selection for remote and local fonts
  * Font loader now picks dynamic fallback weights when requested weights are missing

* **Tests**
  * Added unit tests covering typography sync behavior and font-loading fallback cases

* **Style**
  * Minor editor UI spacing and icon sizing tweaks


kyrapieterse (approved):
Great work

